Issue #4, opened 2015-12-07
Module dependency calculation does not work correctly.


c-frater:
Title: Module Name: short description

Hexagon version: 1.0.1
Build Config: 
```
{
  dest: 'libs/hexagonjs',
  modules: ['date-picker']
}
```

Expected
--------
Build a version of hexagon including the date picker module and all dependencies

Actual
------
A version including only the date picker module is built

Example
-------
```
var hexagon = require('hexagon-js')
return hexagon.light.build({
  dest: 'libs/hexagonjs',
  modules: ['date-picker']
})
```

PR #5, opened 2015-12-07
add dependency calculation to the build

c-frater:
fix for issue #4


build/main/compile.js
```diff
@@ -322,23 +322,27 @@ function getModulesOrder (modules, allModules) {
 
 function getModuleList (options) {
   var allModules = Object.keys(options.allModules)
-  var obj = {}
   if (options.modules) {
-    allModules.filter(function (moduleName) {
+    moduleList = allModules.filter(function (moduleName) {
```
jmsmyth: missing var?

PR #57, opened 2015-12-20
View enhancements

georgesimms:
Various improvements to the view module

* The default view enter function is slightly cleverer - it can handle multiple classes, for instance
* You get a warning if you try to return an element that doesn't match the view selector
* Tests were added

### DIFF
--- a/modules/view/main/index.coffee
+++ b/modules/view/main/index.coffee
@@ -2,7 +2,12 @@
 class View
   constructor: (@rootSelection, @selector, @defaultType) ->
     self = this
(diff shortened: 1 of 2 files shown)


jmsmyth:
+1

modules/view/main/index.coffee
```diff
@@ -12,71 +17,77 @@ class View
   #XXX: Tidy and optimise
```
c-frater: This comment can probably go? Looks pretty tidy/optimised to me :)

georgesimms: idk, I'm not sure I particularly like the "Let's start with an empty array and push stuff into it" but I didn't want to modify too much without adding more tests

Issue #62, opened 2015-12-21
Tests: get npm test to return a failure code upon failure


jmsmyth:
To get travis reporting failures properly

c-frater:
@jmsmyth Is this done now? Just noticed it's still open...

PR #74, opened 2015-12-31
Set process exit code to result from Karma in tests

andersschuller:
In order to have a non-zero exit code from npm test when there are test failures (see #62).

I am actually expecting this to fail on Travis, since the [.travis.yml](https://github.com/ocadotechnology/hexagonjs/blob/c7707a9232f121698e0b6e3f80586d2061a090ac/.travis.yml) currently includes node 0.10 and 0.11, which I believe do not have support for `process.exitCode`. But since we are already going to restrict the project to [at least 0.12](https://github.com/ocadotechnology/hexagonjs/commit/35c4552b794d579f482adb2b7d43e93d1d27b56d), I thought I would just add another commit later to remove the earlier entries from the Travis configuration.


jmsmyth:
Looks great. Shall we remove 0.10 from the .travis.yml list before this gets merged into master?

andersschuller:
Yep, that was the plan, but I was still testing my changes to see if it covered all the cases. Currently it seems that `npm test` won't fail correctly if the setting up of the tests fails, but it does if an actual test fails.

More specifically, using node 0.10, running
```
npm run test-build; echo "EXIT CODE: $?"
```
gives `EXIT CODE: 1`, but
```
npm run test-library; echo "EXIT CODE: $?"
```
gives `EXIT CODE: 0`.

They both show errors like:

```
Unhandled rejection TypeError: Object [object Object] has no method 'accessAsync'
```

but for `test-library` this seems to be in the set up of the tests, and for whatever reason this error isn't being caught. I've tried to add some more Promise rejection handlers, but so far I haven't been able to catch that error in a way that it will actually fail the process. I'm not sure how much of a problem this is, though.

Issue #70, opened 2015-12-29
New feature: hx.text


georgesimms:
This is like hx.json, but doesn't parse JSON, it just calls back with the raw text.

This would be slightly higher-level than hx.request in the sense that it wouldn't involve handling XMLHttpRequests directly.

PR #80, opened 2016-01-03
Request improvements

georgesimms:
* Introduces an hx.text that passes through the responseText rather than an xhr
* Creates a 'hx.reshapedRequest' that determines which type of parser to use based on the Content-Type header


modules/request/main/index.coffee
```diff
@@ -113,29 +113,33 @@ hx.request = -> # url, callback, options || url, data, callback, options
   hx_xhr urlType, urls, data or null, callback, options
 
 
-reshapedRequest = (args, type, formatter) ->
-  urls = args[0]
+parsers =
+  'application/json': (text) -> if text then JSON.parse text
+  'text/html': (text) -> hx.parseHTML text
+  'text/plain': (text) -> text
```
georgesimms: These parsers operate on the string, not the XHR.

modules/request/main/index.coffee
```diff
@@ -113,29 +113,33 @@ hx.request = -> # url, callback, options || url, data, callback, options
...
-  reshapedRequest arguments, 'application/json', (result) ->
-    if result.responseText then JSON.parse(result.responseText)
-
-hx.html = ->
-  reshapedRequest arguments, 'text/html', (result) ->
-    hx.parseHTML result.responseText
+hx.json = (args...) -> reshapedRequest args, 'application/json'
+hx.html = (args...) -> reshapedRequest args, 'text/html'
+hx.text = (args...) -> reshapedRequest args, 'text/plain'
+hx.reshapedRequest = (args...) -> reshapedRequest args
```
georgesimms: use splat feature of coffeescript

modules/request/main/index.coffee
```diff
@@ -113,29 +113,33 @@ hx.request = -> # url, callback, options || url, data, callback, options
...
+  'text/plain': (text) -> text
 
-  if hx.isFunction(args[1])
-    callback = args[1]
-    options = args[2]
-  else
-    data = args[1]
-    callback = args[2]
-    options = args[3]
+reshapedRequest = ([urls, data, callback, options], type) ->
```
georgesimms: use coffeescript destructuring

modules/request/test/spec.coffee
```diff
@@ -624,7 +624,83 @@ describe 'hx-request', ->
...
+        hx.reshapedRequest 'test.html', cb
+
+        expect(jasmine.Ajax.requests.at(0).method).toEqual('GET')
+        jasmine.Ajax.requests.at(0).respondWith
+          status: 200
+          contentType: 'text/html'
+          responseText: '<div class="parent"><span class="child">bob</span></div>'
+
+    it 'should pass the data through', ->
+      data = {some: 'data'}
```
georgesimms: On a more general note, is it ok that we assume that the post data is always JSON, irrespective of the mime type of the response ? More traditional servers might expect form-encoded data, for instance.

c-frater:
Fixes #70

Issue #23, opened 2015-12-08
All: the timezone used by modules that deal with dates should be configurable


c-frater:
An addition to the api to allow the timezone for any of the modules that deal with dates. This should probably be done in the same way as the feeds for locales are done.

c-frater:
After some thought and discussion, the Timezone api should be something that is part of the date/time picker feed(s) that takes a time offset and modifies the output, in the same way as the locale is applied.

If a 'named' timezone is preferred, this will probably need to be managed externally as including the timezone name/location lookup in hexagon would be quite a bit of overhead for not much gain, especially as there are several libraries that already deal with timezones.

Issue #18, opened 2015-12-08
All: User-facing text should be configurable


c-frater:
It should be possible to define the text used by hexagon when creating elements (such as 'Sort By' and the default form validation messages) to enable the localisation of text

georgesimms:
I've been thinking about this one a bit more. How would you all feel about The ability to supply text for multiple locales?

Something like

      hx.userFacingText('noData', {en: 'No data', fr: 'Aucunes données'})

The getter would then check the current locale using the preferences module and fetch the appropriate text accordingly.

There are some issues here - there needs to be well defined behavior in the case where the string is not defined for the current locale (console warning and fall back to English?)

Also this would break getter-setter symmetry, which might be a bit awkward.

PR #81, opened 2016-01-03
Add support for hexagon listening to its own locale changes

georgesimms:
* Deprecate hx.DatePicker::locale etc - components should get the locale directly from the preferences module
* Make it impossible for locale to be missing - it defaults first to the moment locale if moment is present, then to the browser locale, then to US English.
* Use American spellings for some non-external-facing things

### DIFF
--- a/modules/preferences/test/spec.coffee
+++ b/modules/preferences/test/spec.coffee
@@ -5,6 +5,14 @@ describe 'hx-preferences', ->
       expect(hx.preferences.supportedLocales(list)).toEqual(hx.preferences)
       expect(hx.preferences.supportedLocales()).toEqual(list)
 
+    it 'locale: should not be possible to explicitly clear the locale', ->
+      # sanity check
+      expect(hx.preferences.locale()).toBeDefined()
+
+      hx.preferences.locale undefined
+      expect(hx.preferences.locale()).toBeDefined()
+
+
     it 'locale: setter/getter', ->
       expect(hx.preferences.locale('vi')).toEqual(hx.preferences)
       expect(hx.preferences.locale()).toEqual('vi')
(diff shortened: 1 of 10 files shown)


georgesimms:
I considered additionally moving Localizer and LocalizerMoment out into their own module to reduce copypasta between DatePicker and TimePicket. We may want to do this, and add an IntlLocalizer, as part of issue #18

c-frater:
I had a look at merging the localisers and moving them to a separate module and I think it makes a lot of sense and agree that it makes sense to do it as part of #18. It may also make sense to do it as part of #23 as I expect that the changes for the timezone will need their own module too. 

Also, please reference the issue number in pull requests so that it's easy to tell when something you've changed is relevant to an issue.

modules/preferences/test/spec.coffee
```diff
@@ -5,6 +5,14 @@ describe 'hx-preferences', ->
       expect(hx.preferences.supportedLocales(list)).toEqual(hx.preferences)
       expect(hx.preferences.supportedLocales()).toEqual(list)
 
+    it 'locale: should not be possible to explicitly clear the locale', ->
+      # sanity check
+      expect(hx.preferences.locale()).toBeDefined()
+
+      hx.preferences.locale undefined
+      expect(hx.preferences.locale()).toBeDefined()
```
c-frater: Should this not be set back to the default locale (e.g. `navigator.language`) rather than seemingly doing nothing?

c-frater: Discussed this with @jmsmyth and he agrees that rather than doing nothing, this should use the `navigator.language`
meaning that the following is true:
```
hx.preferences.locale(undefined)
hx.preferences.locale().should.equal(navigator.language)
```

PR #75, opened 2015-12-31
Print Mode

c-frater:
This pull request should fix #19

Adds a new module called 'print' that contains the print styles (which are overrides for some of the styles as well as helper styles for making the print display nicely)

This needs more thought regarding how to 'override' themes and whether or not it makes sense to do it this way.

This also needs to be merged into 1.1.0 before the branch is merged

PR #87, opened 2016-01-06
Merge changes from 1.1.0

c-frater:
This should probably be merged before #75


modules/data-table/main/index.coffee
```diff
@@ -209,12 +209,27 @@ class DataTable extends hx.EventEmitter
...
+        if arguments.length > 1
+          options.columns[columnId] ?= {}
+          options.columns[columnId][name] = value
+          @emit(name.toLowerCase() + 'change', {column: columnId, value: value, cause: 'api'})
+          @render(cb)
+          this
+        else if options.columns[columnId]
+          options.columns[columnId][name]
+
+  maxWidth: columnOnlyOption('maxWidth')
```
jmsmyth: Is there a reason this can't be a normal columnOption?

c-frater: I guess it could be but I didn't think it made sense to let you set the maxWidth for every column (especially as it doesn't seem to do anything if you set a value for every column)

package.json
```diff
@@ -1,11 +1,14 @@
 {
   "name": "hexagon-js",
-  "version": "1.0.4",
+  "version": "1.1.0",
```
jmsmyth: Should probably only be bumped when the actual release happens?

c-frater:
@jmsmyth Is this ok to merge now?

jmsmyth:
@c-frater Go for it :-)

Issue #24, opened 2015-12-08
Preferences/All: Hexagon should be able to listen to it's own locale changes


c-frater:
Currently, the preferences object allows the locale to be changed but this has to be manually propagated through elements such as the date picker. It would be much better if the modules that deal with locales could listen to locale changes and update automatically.

PR #90, opened 2016-01-07
Timezone

c-frater:
Merged the date picker and time picker feeds into a new module (date localizer)
Added support for timezones to preferences

This is probably in need of a more in-depth review as I'm not sure if there's a better way to implement the functionality

This is for #23 and partly for #24


modules/date-localizer/main/index.coffee
```diff
@@ -0,0 +1,168 @@
+class DateLocalizer
```
georgesimms: +1 for making this a separate module

modules/date-localizer/main/index.coffee
```diff
@@ -0,0 +1,168 @@
...
+    if useInbuilt
+      order = ['YYYY', 'MM', 'DD']
+      split = dateString.split('-')
+    else
+      order = @dateOrder()
+      split = dateString.split('/')
+    allValid = split.length is 3 and not split.some (e) -> e is '' or e is '0'
+    if allValid
+      format = ''
+      for w, i in order
```
georgesimms: or:

    hx.zip [w, split]
      .map ([w, splitElemStr]) ->
        splitElem = Number splitElemStr
        switch w
          when 'DD'
            daysValid = part.length < 3 and part isnt ''
            day = splitElem
       # etc

modules/date-localizer/main/index.coffee
```diff
@@ -0,0 +1,168 @@
...
+    moment({year: year}).locale(hx.preferences.locale()).format('YYYY')
+
+  decade: (start, end) ->
+    @year(start) + ' - ' + @year(end)
+
+  date: (date) ->
+    moment(date).locale(hx.preferences.locale()).format('L')
+
+  time: (date, showSeconds) ->
+    date = hx.preferences.applyTimezoneOffset(date)
```
georgesimms: +1 for getting the time-zone-config via the preferences module

modules/date-localizer/test/spec.coffee
```diff
@@ -0,0 +1,81 @@
...
+
+  it 'checkTime: should return true when time is valid', ->
+    localizer.checkTime('1:30:00'.split(':')).should.equal(true)
+
+  it 'checkTime: should return false when time is not valid', ->
+    localizer.checkTime('a:30'.split(':')).should.equal(false)
+    localizer.checkTime('1:b:00'.split(':')).should.equal(false)
+
+  it 'stringToDate: should convert a localized date string back to a date object', ->
+    localizer.stringToDate('07/01/2016').should.eql(zeroHourDate)
```
georgesimms: What does eql do ?

modules/date-localizer/main/index.coffee
georgesimms: Somewhat contrary to its name, this localizes both Dates and Times. Might it be worth calling it DateTimeLocalizer or Localizer or similar ? (Related to #18, but you might want to do that in a separate module)

modules/date-localizer/test/spec.coffee
c-frater: It's a chai helper for deep equal, useful when comparing date objects directly (without having to use 'tostring' or 'gettime')

As a note, tests should be using chai rather than jasmine - we're phasing it out

modules/date-localizer/main/index.coffee
c-frater: I guess it does make sense - the reason I chose `DateLocalizer` is because it was shorter and technically it does only localize date objects - one of the formats happens to be time.

I'm fine with changing it to `DateTimeLocalizer`

c-frater:
Tests failing due to #97 which is unrelated to the changes made in this merge request

modules/date-localizer/test/spec.coffee
```diff
@@ -0,0 +1,81 @@
+describe 'dateTimeLocalizerMoment', ->
+  # Not sure how to test this...
```
jmsmyth: This could allow passing in the moment object/function for testing purposes - then once everything is using webpack it can require('moment') and pass it in here

jmsmyth:
+1

Issue #85, opened 2016-01-05
Data Table: weird behavior with deselecting rows as a user that were set by the API


georgesimms:
Consider the following

    const feed = hx.dataTable.objectFeed({
      headers: [
        { name: 'Name', id: 'name' },
        { name: 'Age', id: 'age' },
        { name: 'Profession', id: 'profession' }
      ],
      rows: [
         {
           id: 0, // hidden details can go here (not in the cells object)
           cells: { 'name': 'Bob', 'age': 25, 'profession': 'Developer' }
         },
         {
           id: 1,
           cells: { 'name': 'Jan', 'age': 41, 'profession': 'Artist' }
         },
         {
           id: 2,
           cells: { 'name': 'Dan', 'age': 41, 'profession': 'Builder' }
         }
      ]
    })
    const tableSel = hx.detached('div')
    const table2 = new hx.DataTable(tableSel.node(), {feed: feed, singleSelection: true, selectEnabled: true})
    table2.selectedRows([0])
    table2.on('selectedrowschange', (data) => data.cause === 'user' ? console.log(data.value) : undefined)
    hx.select('#example').append(tableSel)

Everything seems fine ?

Now try deselecting a row by clicking on it!

It doesn't get deselected.

Also `[0]` is logged, I would expect `[]`.

Clicking it a second time "works", so all the user sees is a slow page.

c-frater:
Is this resolved by your previous merge request #67?

georgesimms:
I haven't checked

georgesimms:
I wrote a regression test for this. Unfortunately it still fails :(

georgesimms:
Fixed by #98

PR #96, opened 2016-01-08
Add regression test for unselecting a row selected by the API

georgesimms:
A regression test for #85 . Unfortunately the underlying bug is still present :-(


georgesimms:
bug fixed by #98

PR #98, opened 2016-01-09
Fix unselecting api-selected row bug

### LINKED ISSUE #85: Data Table: weird behavior with deselecting rows as a user that were set by the API

georgesimms:
fixes #85 . The issue is selectRow clears the selection if @_lastSelected is different from the row you want to select and single selection is enabled, but @_lastSelected isn't set by hx.DataTable::selectedRows.

### DIFF
--- a/modules/data-table/main/index.coffee
+++ b/modules/data-table/main/index.coffee
@@ -287,7 +287,9 @@ class DataTable extends hx.EventEmitter
       else
         options[name]
 
-    rowToArray = (headers, obj) -> headers.map (header) -> obj.cells[header.id]
+    rowToArray = (headers, obj) -> headers.map (header) ->
+      console.log obj unless obj.cells
+      obj.cells[header.id]
 
     # build the main structure of the table in a detached container
     container = hx.detached('div').class('hx-data-table-content')
--- a/modules/data-table/main/index.coffee
+++ b/modules/data-table/main/index.coffee
@@ -287,9 +287,7 @@ class DataTable extends hx.EventEmitter
       else
         options[name]
 
-    rowToArray = (headers, obj) -> headers.map (header) ->
-      console.log obj unless obj.cells
-      obj.cells[header.id]
+    rowToArray = (headers, obj) -> headers.map (header) -> obj.cells[header.id]
 
     # build the main structure of the table in a detached container
     container = hx.detached('div').class('hx-data-table-content')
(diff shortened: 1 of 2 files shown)


modules/data-table/main/index.coffee
```diff
@@ -236,7 +236,9 @@ class DataTable extends hx.EventEmitter
       if @singleSelection() and hx.isArray(value) and value.length
         value = [value[0]]
       @_.selectedRows = new hx.Set(value)
-      @emit('selectedrowschange', {value: @_.selectedRows.values(), cause: 'api'})
+      newSelectedRows = @_.selectedRows.values()
+      @emit('selectedrowschange', {value: newSelectedRows, cause: 'api'})
+      @_.lastSelected = newSelectedRows[newSelectedRows.length - 1]
```
georgesimms: this is the salient line. It sets @_.lastSelected when the selected rows were changed by the API

modules/data-table/test/spec.coffee
```diff
@@ -688,6 +692,35 @@ describe 'data-table', ->
...
+        table.selectedRows [0]
+        
+        table.on 'selectedrowschange', (data) ->
+          if data.cause is 'user'
+            # Row 0 was selected before, so now we're unselecting it
+            data.value.should.eql []
+            done()
+        checkSel = tableSel.select '.hx-sticky-table-wrapper .hx-data-table-checkbox'
+        faker = fakeNodeEvent checkSel.node()
+        faker fakeEvent
```
georgesimms: regression test for bug

PR #99, opened 2016-01-10
Fluent graphs

georgesimms:
Tests now pass.


modules/plot/main/coffee/Expose.coffee
```diff
@@ -7,6 +7,14 @@ hx.plot.arcCurve = arcCurve
 hx.plot.svgCurve = svgCurve
 hx.Axis = Axis
 hx.Graph = Graph
+
+hx.graph = (options) ->
+  selection = new hx.detached 'div'
+  graph = new Graph selection.node(), options
+  # There is no point rendering it now, the selection is of zero size.
+  selection.on 'resize', -> graph.render()
```
georgesimms: We need to do this as the selection will be initially of zero height/width

We might want to think about doing this for all graphs ? Would it be expensive ?

modules/plot/test/spec.coffee
```diff
@@ -243,15 +254,24 @@ describe "plot", ->
...
-      it 'should display when the height is positive', ->
-        selection = hx.detached('div').style('height', '400px')
-        hx.select('body').add(selection)
-        graph = new hx.Graph(selection.node())
-        axis = graph.addAxis('linear', 'linear')
-        a1 = axis.addSeries()
-        graph.render()
-        selection.selectAll('.hx-series').size().should.equal(2)
-        selection.remove()
+      it 'should display when the height is positive', (done) ->
```
georgesimms: This test was added as an example of how to use the fluent API.

modules/plot/test/spec.coffee
```diff
@@ -243,15 +254,24 @@ describe "plot", ->
...
+        graphOpts =
+          axes: [
+            series: [['line']]
```
georgesimms: The series option for an axis and axes option for a graph also exist for graphs constructed with a constructor.

modules/plot/main/coffee/Graph.coffee
```diff
@@ -294,15 +297,16 @@ class Graph extends hx.EventEmitter
       .attr('width', @plotArea.x2 - @plotArea.x1)
       .attr('height', @plotArea.y2 - @plotArea.y1)
 
+    @emit 'render'
```
georgesimms: This event was added mostly to facilitate testing, but you might want to document it anyway.

modules/plot/main/coffee/Expose.coffee
c-frater: Plot already adds a resize listener on line 28 of Graph.coffee in the hx.Graph constructor so I don't think the on-resize is needed?
```
selection = hx.select(@selector).on 'resize', 'hx.plot', => @render()
```

Also, when adding event listeners inside hexagon, you should namespace them in the format `hx.{module-name}`, e.g. `hx.plot`, `hx.data-table` etc

modules/plot/test/spec.coffee
```diff
@@ -243,15 +254,24 @@ describe "plot", ->
       checkOptionAndSetterGetter(hx.BarSeries, 'group', ['group1', 'group2'])
 
     describe "render", ->
-      it 'should display when the height is positive', ->
```
c-frater: I think this test should probably be left in as it's testing the existing method for creating graphs

modules/plot/test/spec.coffee
```diff
@@ -243,15 +254,24 @@ describe "plot", ->
...
+        graphOpts =
+          axes: [
+            series: [['line']]
+            x:
+              type: 'linear'
+            y:
+              type: 'linear'
+          ]
+        selection = hx.graph graphOpts
+          .style 'height', '400px'
```
c-frater: If you set this height after appending to the body it might resolve the failing test issue.

modules/plot/main/coffee/Expose.coffee
```diff
@@ -7,6 +7,14 @@ hx.plot.arcCurve = arcCurve
 hx.plot.svgCurve = svgCurve
 hx.Axis = Axis
 hx.Graph = Graph
+
+hx.graph = (options) ->
+  selection = new hx.detached 'div'
+  graph = new Graph selection.node(), options
+  # There is no point rendering it now, the selection is of zero size.
+  selection.on 'resize', 'hx.plot', -> graph.render()
```
c-frater: This line isn't needed. the hx.Graph constructor already does this.

The previous comment was more for future reference about namespacing events

modules/plot/test/spec.coffee
c-frater: Looking at the other test, I definitely think this test should be left in.

The newly added test is checking that the fluent api works, it might be worth adding further test(s) to make sure the fluid graph api works.

modules/plot/main/coffee/Axis.coffee
```diff
@@ -70,8 +70,8 @@ class Axis
       series: new hx.List
     }
 
-    @x = dimension(this, opts.x)
-    @y = dimension(this, opts.y)
+    @x = dimension this, hx.merge axisTickLabelPosition: 'bottom', options?.x
```
jmsmyth: Might be clearer formatted like this:

    hx.merge({
      axisTickLabelPosition: 'bottom'
    }, options.x)

modules/plot/main/coffee/Axis.coffee
```diff
@@ -85,6 +85,8 @@ class Axis
     @xTitleHeight = 0
     @yAxisSize = 50
     @yTitleHeight = 0
+    options?.series?.forEach ([seriesType, seriesOpts]) =>
```
jmsmyth: Looks like series have to be provided as tuples? Perhaps objects would make more sense?
```
options.series = [
  {
    type: 'line',
    options: {...}
  },
  {
    type: 'bar',
    options: {...}
  }
]
```

c-frater: +1

demo/index.html
```diff
@@ -1380,21 +1370,30 @@
...
-                labelInterpolated: true,
-                markersEnabled: true,
-                strokeEnabled: true,
-                strokeColor: col,
-                fillEnabled: true,
-                fillColor: hx.color(col).alpha(0.2).toString(),
-                group: 'some-group'
-              })
-            })
+            const axes = [{
```
c-frater: Does using const here have any implications? The rest of the page was not written with ES6 in mind so might be worth using `var` here

c-frater:
Once changes have been made to axes to use objects instead of tuples and other comments addressed, this should be fine to merge.
```
options.series = [
  {
    type: 'line',
    options: {...}
  },
  {
    type: 'bar',
    options: {...}
  }
]
```

modules/plot/main/coffee/Axis.coffee
georgesimms: done

demo/index.html
georgesimms: done

c-frater:
I'm happy with this (once the tests pass) - ready to merge @jmsmyth?

PR #106, opened 2016-01-14
Update karma config

c-frater:
The intermittent failing tests may be due to leftover elements in the DOM between tests. This should help alleviate the issue


c-frater:
The tests did pass however opening/closing the branch accidentally kicked off the tests again and then caused them to fail

Issue #104, opened 2016-01-13
Selection: add an api for selecting from the direct children of an element


c-frater:
It's often useful to be able to select from direct children.

Adding an option to hx.select to allow shallow selection or adding a new method (e.g. `shallowSelect`) would accomplish this.

c-frater:
Moved into this milestone as the number of issues that would either be much easier to solve with this addition is increasing

PR #107, opened 2016-01-14
fixes for nested sticky table headers

c-frater:


modules/sticky-table-headers/main/index.coffee
```diff
@@ -43,6 +43,21 @@ cloneEvents = (elem, clone) ->
...
+getChildrenFromTable = (t, body, single) ->
+  realParents = getChildren(t.select(if body then 'tbody' else 'thead'), 'tr')
+  hx.flatten realParents.map (parent) -> getChildren(hx.select(parent), 'th, td', single)
+
+getChildren = (parent, selector, single) ->
+  children = if single
+    parent.select selector
+  else
+    parent.selectAll selector
+  children.filter((node) -> node.node().parentNode is parent.node()).nodes
```
c-frater: `getChildren` is a basic implementation of logic that could be used for #104 - although I'm not sure how useful this is

modules/table/main/index.scss
```diff
@@ -25,7 +25,7 @@
     }
   }
 
-  tbody th:first-child {
+  tbody > tr > th:first-child {
```
c-frater: Prevents nested tables getting incorrect borders

modules/sticky-table-headers/main/index.coffee
```diff
@@ -43,6 +43,21 @@ cloneEvents = (elem, clone) ->
       cloneChildren = clone.childNodes
       cloneEvents elemChildren[i], cloneChildren[i] for i in [0..elemChildren.length]
 
+getChildrenFromTable = (t, body, single) ->
+  realParents = getChildren(t.select(if body then 'tbody' else 'thead'), 'tr')
+  hx.flatten realParents.map (parent) -> getChildren(hx.select(parent), 'th, td', single)
+
+getChildren = (parent, selector, single) ->
+  children = if single
```
jmsmyth: If `parent.children` is available it could use that

modules/sticky-table-headers/main/index.coffee
```diff
@@ -194,6 +212,7 @@ class StickyTableHeaders
 
     if options.fullWidth
       table
+        .style('width', '')
```
jmsmyth: .style('width', undefined)

jmsmyth:
+1

modules/sticky-table-headers/main/index.coffee
c-frater: Avoided that as `parent.children` is a `NodeList`, not an `Array` and using children would also imply use of `Element.matches` which doesn't work in phantom. 
It's easier to select everything that matches the selector and filter than it is to check if a node matches a selector in this case. 
I'd also assume that filtering on a selection and comparing nodes would be quicker than using `Element.matches` on all the direct children.

jmsmyth: Ah yes, I forgot about phantom.

I suppose since you are only checking the nodeType you _could_ get away with just filtering on `node.nodeName`, something like:

```
  children = Array.prototype.slice.call(parent.children).filter(node => node.nodeName === 'TBODY')
```

but it probably isn't worth changing now.

Its hard to tell if the last assumption would necessarily be the case without benchmarking it - my initial assumption was the opposite :-p 
But, this is something that can be investigated for #104 (ie best performance characteristics and best browser support). It probably depends a lot on what you are selecting

modules/sticky-table-headers/main/index.coffee
```diff
@@ -258,28 +274,33 @@ class StickyTableHeaders
...
-        hx.select(realNodes[i]).classed('hx-sticky-table-invisible', true)
+      realNodes = getChildrenFromTable table, false, true
+      clonedNodes = getChildrenFromTable topLeftTable, false, true
 
+      createStickyHeaderNodes realNodes, clonedNodes
 
     topHead?.style('height', offsetHeight + 'px')
       .style('width', wrapperBox.width + 'px')
       .style('left', offsetWidth + 'px')
+      .selectAll('.hx-sticky-table-invisible')
```
jmsmyth: not quite sure what this is doing, but will this selection be okay if there are nested tables?

jmsmyth: Actually looking at this again, it might be okay - since it selecting on the thead, is that right?

c-frater: The `hx-sticky-table-invisible` class is added to the headers (left, top, top-left) in the main table so they are invisible but still affect the layout (`visibility: hidden`). 
The class isn't needed inside the cloned tables and is removed here as otherwise it causes problems with nested tables due to the styles:
```
.hx-sticky-table-wrapper .hx-sticky-table-invisible {
    visibility: hidden;
}  
```

It's a lot easier to remove the class than add nesting to the css.

Also, the class is added in `createStickyHeaderNodes` (line 60) after the clone has been made, however the class applied to the original is inherited by the clone and removing the class from the clone didn't work, hence why this removal is required.

c-frater: Just saw your second comment. 
topHead is a clone of the table with a 'mask' applied so you only see the header, leftHead is the same but for the first column, topLeftHead is the same but for the top left cell.
They are complete clones as it's the only way to guarantee the tables all have the same column/row widths and heights without having to fix the sizes for every cell.
The three elements are added by sticky table headers and shouldn't have any nested functionality so classing here shouldn't be a problem.

PR #123, opened 2016-01-31
Specrunner

georgesimms:
Adds html files to run tests in the browser. This is useful for visually / interactively debugging things using your favorite browser's debugger, and enables a quick feedback loop.

It also allows testing in browsers that might not have karma support (eg edge).

It requires some tests to be modified to not use the entire body but only a selection with in it called .fixture.

Also I found lots of references to sinon everywhere but we don't seem to be using it. Can it be dropped?


build/main/builder.js
```diff
@@ -106,6 +104,7 @@ function getModuleList (moduleDirectories) {
     })
     .then(function (moduleFolders) {
       var moduleDirectoryMap = {}
+      var i, j
```
georgesimms: otherwise i and j are globals

modules/data-table/test/spec.coffee
```diff
@@ -445,7 +446,7 @@ describe 'data-table', ->
           testTable {tableOptions: {displayMode: 'paginate', pageSize: 1}}, undefined, (container, dt, options, data) ->
             dt.on 'render', ->
               dt.page().should.equal(2)
-              hx.select('body').clear()
```
georgesimms: Don't delete the body ! (brings down the specrunner with it)

specrunner.jade
```diff
@@ -0,0 +1,14 @@
+html
```
georgesimms: Perhaps, something like this can be done with UM. But I don't know.

build/main/test.js
```diff
@@ -110,12 +111,96 @@ function runTests (moduleNames, destDir, phantomOnly) {
...
+    'chai-spies.js'
+  ))
+  var dependencies = Promise.all([jasmine, chai, chaiSpies])
   return buildTestPackages(moduleNames)
     .then(function (testPackages) {
-      // XXX: build a html page with test embedded + write to the destFilename
+      var promises = testPackages.map(function (testPackage) {
+        return getOutputFileNames(destDir, testPackage)
+      })
+      // XXX zip here
```
georgesimms: Is there a node-js equivalent of `hx.zip`?

build/main/specrunner.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+function watch (files, moduleName) {
+  gaze(files, function (error, watcher) {
+    if (error) {
+      throw error
+    }
+    watcher.on('changed', function () {
+      console.log('Rebuilding ' + moduleName)
+      build([moduleName])
```
georgesimms: Only rebuild the edited module.

build/main/test.js
jmsmyth: [zip-array](https://www.npmjs.com/package/zip-array) or [array-zip](https://www.npmjs.com/package/array-zip)?

c-frater:
This will need to be updated to work with #132

c-frater:
This is no longer relevant - closing.

Issue #124, opened 2016-02-02
Convert tests over to use chai


jmsmyth:
- [x] animate
- [x] click-detector
- [x] collapsible
- [x] color
- [x] date-localizer
- [x] data-table
- [x] date-picker
- [x] date-time-picker
- [x] drag-container
- [x] drawing
- [x] dropdown (in progress)
- [x] event-emitter
- [x] filter
- [x] format
- [x] interpolate
- [x] list
- [x] map
- [x] meter
- [x] palette
- [x] plot
- [x] preferences
- [x] progress-bar
- [x] request (in progress)
- [x] selection
- [x] set
- [x] slider
- [x] sort
- [x] tag-input
- [x] time-picker
- [x] time-slider
- [x] titlebar
- [x] transition
- [x] util
- [x] view

PR #126, opened 2016-02-02
Use chai for tests

c-frater:
#124 - probably needs some sense-checking (mostly related to spies) but tests all pass and use chai syntax

### DIFF
--- a/modules/animate/test/spec.coffee
+++ b/modules/animate/test/spec.coffee
@@ -28,8 +28,11 @@ describe 'hx-animate', ->
 
   savedGetComputedStyle = undefined
   savedHxLoop = undefined
+  clock = undefined
+  beforeAll -> clock = sinon.useFakeTimers()
+  afterAll -> clock.restore()
+
   beforeEach ->
-    jasmine.clock().install()
     savedGetComputedStyle = window.getComputedStyle
     window.getComputedStyle = (node, thing) ->
       {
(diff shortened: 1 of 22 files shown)


modules/animate/test/spec.coffee
```diff
@@ -927,20 +928,20 @@ describe 'hx-animate', ->
...
     it 'a warning should be thrown when a named morph is used that doesnt exist', ->
-      chai.spy.on(console, 'warn')
+      hx.consoleWarning = chai.spy()
 
       hx.morph(new FakeNode)
         .with('delay5', 500)
         .go()
 
-      console.warn.should.have.been.called.once
+      hx.consoleWarning.should.have.been.called.once
```
jmsmyth: `hx.consoleWarning` should be reset back to its original at the end of the test

c-frater: done :)

modules/request/main/index.coffee
```diff
@@ -1,13 +1,9 @@
-hasResponse = (request) ->
-  responseType = request.responseType
-  if responseType and responseType isnt 'text' then request.response else request.responseText
```
c-frater: This file probably needs looking over just to make sure it's still sensible. The changes made are:
- Remove support for old browsers (XDomainRequest and onreadystatechange)
- Remove support for request responses that have content but no status as this should not be possible
- Restructure reshaped request to output less code (and make 100% coverage possible)

Issue #100, opened 2016-01-12
Dropdown: update the tests


c-frater:
The tests are currently buggy and several fail intermittently, causing the build to fail.
They could do with being re-factored to use events rather than `jasmine.clock` and the clean-up between the tests seems to need work as well.

jmsmyth:
resolved by #128

PR #128, opened 2016-02-03
Fix the dropdown tests

jmsmyth:
This change does a couple of things:
 - Moves all things in dropdown that are not documented onto an underscore object
 - Separates the calculation of the dropdown position from the dom querying and dom manipulation (which will hopefully make the tests not fail occasionally)
 - Converts the tests over to use chai

### DIFF
--- a/modules/dropdown/main/theme.scss
+++ b/modules/dropdown/main/theme.scss
@@ -2,4 +2,5 @@
   background: $background-col;
   box-shadow: 0px 2px 2px 2px $shadow-col;
   border-color: $border-col;
+  margin-top: $spacing * 1px;
 }
--- a/modules/dropdown/test/spec.coffee
+++ b/modules/dropdown/test/spec.coffee
@@ -81,7 +81,7 @@ describe 'hx-dropdown', ->
     button = makeButton()
 
   afterEach ->
-    hx.component(id).cleanUp()
+    hx.component(id)?.cleanUp()
     hx.selectAll('.hx-dropdown').remove()
     button = undefined
     fixture.clear()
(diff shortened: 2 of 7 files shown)


c-frater:
I'm guessing this resolves #100?

modules/dropdown/test/spec.coffee
```diff
@@ -61,14 +58,13 @@ describe 'hx-dropdown', ->
     hx.loop = hx_loop = (f) ->
       g = -> hx_loop_update(f, g)
       hx_loop_update(f, g)
-    jasmine.clock().install()
     baseTime = new Date(2013, 0, 1)
-    jasmine.clock().mockDate(baseTime)
+    clock = sinon.useFakeTimers(baseTime.getTime())
```
c-frater: Is setting the baseTime actually required here? It shouldn't matter when the clock runs and `useFakeTimers` seems to default to 0 rather than 'now' so already has a consistent start point

modules/dropdown/main/theme.scss
```diff
@@ -2,4 +2,5 @@
   background: $background-col;
   box-shadow: 0px 2px 2px 2px $shadow-col;
   border-color: $border-col;
+  margin-top: $spacing * 1px;
```
c-frater: So the spacing is applied as part of the styles as well as part of the javascript separately?

modules/menu/main/index.coffee
```diff
@@ -31,7 +31,7 @@ getAllItems = (menu) ->
 # If setActive is called on click, the dropdown is already hidden so
 # offsetParent checking doesn't work.
 setActive = (menu, pos, up, click) ->
-  menu.dropdown.dropdown?.selectAll('.hx-menu-item').classed('hx-menu-active',false)
+  menu.dropdown._.dropdown?.selectAll('.hx-menu-item').classed('hx-menu-active',false)
```
c-frater: I think `dropdown.dropdown` may be used in more places than autocomplete/menu. A quick search brought up inline-picker but there may be more

modules/dropdown/main/theme.scss
jmsmyth: Yeah, default spacing in the styles, then it overrides it if provided in the options when creating the dropdown. Kept the calculation of the dropdown position a bit simpler

c-frater:
Looks good - if you're happy then I think it's ready to merge

PR #135, opened 2016-02-08
data table selectedRows fix

c-frater:
The code used `lastSelected` as a variable which was incorrectly set to the row id in #98 as the variable name is ambiguous.
This resolves the issue raised in #85 and updates the variable name to be less ambiguous


jmsmyth:
Looks good

PR #155, opened 2016-03-04
Data Table URL Feed Fix

### LINKED ISSUE #154: Data Table: Url feed does not work
`hx.dataTable.urlFeed(url)` does not work correctly as it is expecting hx.json to call the callback with `(response)` however in reality it calls it with `(error, response)`

c-frater:
Update the data table url feed so that it uses the correct callback format
Fixes #154

### DIFF
--- a/modules/data-table/main/index.coffee
+++ b/modules/data-table/main/index.coffee
@@ -786,10 +786,18 @@ urlFeed = (url, options) ->
 
   {
     url: url # for debugging
-    headers: maybeCached (r) -> hx.json(url, { type: 'headers', extra: options.extra }, r)
-    totalCount: maybeCached (r) -> hx.json(url, { type: 'totalCount', extra: options.extra }, (res) -> r(res.count))
-    rows: (range, cb) -> hx.json(url, { type: 'rows', range: range, extra: options.extra }, cb)
-    rowsForIds: (ids, lookupRow, cb) -> hx.json(url, { type: 'rowsForIds', ids: ids, extra: options.extra }, cb)
+    headers: maybeCached (cb) ->
+      hx.json url, { type: 'headers', extra: options.extra }, (err, res) ->
+        cb(res)
+    totalCount: maybeCached (cb) ->
+      hx.json url, { type: 'totalCount', extra: options.extra }, (err, res) ->
+        cb(res.count)
+    rows: (range, cb) ->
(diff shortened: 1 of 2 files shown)


modules/data-table/main/index.coffee
```diff
@@ -786,10 +786,18 @@ urlFeed = (url, options) ->
 
   {
     url: url # for debugging
-    headers: maybeCached (r) -> hx.json(url, { type: 'headers', extra: options.extra }, r)
-    totalCount: maybeCached (r) -> hx.json(url, { type: 'totalCount', extra: options.extra }, (res) -> r(res.count))
-    rows: (range, cb) -> hx.json(url, { type: 'rows', range: range, extra: options.extra }, cb)
-    rowsForIds: (ids, lookupRow, cb) -> hx.json(url, { type: 'rowsForIds', ids: ids, extra: options.extra }, cb)
+    headers: maybeCached (cb) ->
+      hx.json url, { type: 'headers', extra: options.extra }, (err, res) ->
+        cb(res)
```
georgesimms: This discards err without logging it! I'm not sure I like this.

I see a number of options -

     console.error(er) if er 
     throw er if er # might propogate to top of stack trace, or have other weird unexpected results
     hx.notify.negative("Oops! Something went wrong.") # very intrusive, although not unlike what the preference dialogue does.

c-frater: An alternative would be to write your own url feed that has error handling but it makes sense to add it here - I didn't really think about it because it wasn't doing it before but I'll add it in

c-frater: I think `console.error(err)` makes the most sense in this situation in terms of what browsers do when a request has an error. If you need more advanced error handling then you can always write a custom feed.

modules/data-table/main/index.coffee
```diff
@@ -784,12 +784,22 @@ urlFeed = (url, options) ->
     else
       (cb) -> fetcher(cb)
 
+  jsonCallback = (cb) ->
+    (err, value) ->
+      console.error(err) if err
+      cb(value)
```
c-frater: Added this to reduce duplication

jmsmyth:
Looks good - in the long term we might want to add better error handling for the data table feeds so that errors can be displayed on the table somehow. But this is good for now :+1:

Issue #108, opened 2016-01-15
Picker: should add the hx-btn class to the node supplied in the constructor


georgesimms:
In particular, the result of `hx.picker` doesn't have the class (but I think it would make more sense to do it in `hx.Picker` not `hx.picker` because the user should not care)

jmsmyth:
+1

It might also make sense for hx.Picker to check that the selector passed in is for a button element

c-frater:
Had a look at this, I'm not sure it does make sense for hx.Picker to check the selector as you can use `hx-btn` on non-button elements (e.g. span, div etc)

PR #156, opened 2016-03-04
Form builder

### LINKED ISSUE #112: Form Builder: properties can only be disabled when creating and cannot be hidden
When creating the form properties can be disabled but cannot then be easily (and consistently) toggled.
Also, form properties cannot be initialised as hidden but can be toggled later using `hidden`

Disabled and hidden should be consistent. You should be able to initialise properties as disabled/hidden as well as toggle the disabled/hidden state using a method.

c-frater:
Update form builder to fix #112 as well as fixing #108 and adding a fix for menu dropdowns

### DIFF
--- a/modules/base/main/index.scss
+++ b/modules/base/main/index.scss
@@ -51,11 +51,6 @@ input {
   border-width: 1px;
   padding: 0.4em;
 
-  &:disabled {
-    cursor: not-allowed;
-    opacity: 0.5;
-  }
-
   &:invalid {
     border-width: 1px;
     border-style: dotted;
@@ -184,4 +179,13 @@ input[type=radio] {
 
 hr {
   border: 1px solid transparent;
+}
+
+input,
+textarea,
+select {
+  &:disabled {
+    cursor: not-allowed;
+    opacity: 0.5;
+  }
 }
--- a/modules/form-builder/main/index.coffee
+++ b/modules/form-builder/main/index.coffee
@@ -22,7 +22,7 @@ class Form extends hx.EventEmitter
 
     # Define the default function for enabling/disabling a form property
     extras.disable ?= (sel, disable) ->
-      sel.attr('disabled', if disable then true else undefined)
+      sel.attr('disabled', if disable then 'disabled' else undefined)
 
     if extras.key?
       key = extras.key
--- a/modules/form-builder/main/index.coffee
+++ b/modules/form-builder/main/index.coffee
@@ -2,7 +2,6 @@
 class Form extends hx.EventEmitter
   constructor: (@selector) ->
     super
-
     hx.component.register(@selector, this)
 
     @formId = "form-"+hx.randomId() + '-'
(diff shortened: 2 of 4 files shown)


modules/base/main/index.scss
```diff
@@ -184,4 +179,13 @@ input[type=radio] {
...
   border: 1px solid transparent;
+}
+
+input,
+textarea,
+select {
+  &:disabled {
+    cursor: not-allowed;
+    opacity: 0.5;
+  }
```
c-frater: Moved this to the bottom so that the styles aren't overriden for checkbox/radio inputs

modules/form-builder/main/index.coffee
```diff
@@ -21,6 +20,10 @@ class Form extends hx.EventEmitter
     selection = entry.append(nodeType).attr("id",id)
     extras = f.call(selection) or {}
 
+    # Define the default function for enabling/disabling a form property
+    extras.disable ?= (sel, disable) ->
+      sel.attr('disabled', if disable then true else undefined)
```
georgesimms: Maybe `if disable then "true" else undefined`.

c-frater: Not sure what the quotes add apart from two extra characters? It still resolves to `disabled="true"` on the element.
 
Also, I'd normally put `if disable then '' else undefined` because the disabled attribute works on existence not value (e.g. if it has the disabled attribute, it's disabled regardless of the value added) but I thought having `true` was more explicit as to what this is doing.

jmsmyth: Or `if disable then 'disabled' else undefined`

modules/form-builder/main/index.coffee
```diff
@@ -21,6 +20,10 @@ class Form extends hx.EventEmitter
     selection = entry.append(nodeType).attr("id",id)
     extras = f.call(selection) or {}
 
+    # Define the default function for enabling/disabling a form property
+    extras.disable ?= (sel, disable) ->
```
jmsmyth: Is this what is exposed in the public api? If so, can it be made a getter setter?

jmsmyth: Ah, just seen the disabled method below - ignore this :)

modules/form-builder/main/index.coffee
```diff
@@ -262,6 +281,21 @@ class Form extends hx.EventEmitter
       else
         !!prop.hidden
 
+  disabled: (property, disabled) ->
+    if hx.isArray(property)
+      res = property.map (prop) => @disabled(prop, disabled)
+      if disabled? then this else res
```
jmsmyth: Should check arguments.length rather than relying on the value of disabled

jmsmyth: Also, does it need to support property being an array?

c-frater: I just went with what the hidden function already did and that supports an array - It's pretty much the same function so if it needs changing then they both do.

jmsmyth: Okay, if hidden already does it then keep it consistent

Issue #148, opened 2016-02-23
Autocomplete / View: autocomplete is slow when returning large datasets


AireL:
Tried using it with a large dataset (5000+) - when returning a significant chunk of that data the page would lock up for seconds while trying to render the page

c-frater:
This looks to be a problem related to `hx.View` and #14 (fixed in #57) which has [added a check to the view](https://github.com/ocadotechnology/hexagonjs/blob/master/modules/view/main/index.coffee#L75-L78), slowing it down. Is there any way to improve performance for the check @jmsmyth @georgesimms ?

jmsmyth:
The `selectAll` can be avoided - checking the `element.nodeName` and `classes` of the new element match the selector and checking that `rootSelection.node().contains(element)` could be faster.

Also the new function can be optimised a bit [on this line](https://github.com/ocadotechnology/hexagonjs/blob/master/modules/view/main/index.coffee#L6) by moving the deconstruction of the selector out of the new function

Not sure if that will be enough though, since 5000+ is quite a lot of nodes. Perhaps just showing the top n results (where n is a smallish number) could work. If not, someone might need to implement #17?

georgesimms:
There's an implicit suggestion here that the view selection is relatively simple (like `element-type.class-1.class-2` or something). This is OK I guess but we should be honest about it.

Also we may want to disable the warning when the default view enter function is used.

PR #157, opened 2016-03-07
Increase the performance of View

c-frater:
fix for #148 

The below test case shows an autocomplete with 10000 items which renders in around 1 second, as opposed to previously when it rendered in around 30-40 seconds.
```html
<input id="test" />
<script>
  new hx.AutoComplete('#test', hx.range(10000))
</script>
```

### DIFF
--- a/modules/menu/main/index.coffee
+++ b/modules/menu/main/index.coffee
@@ -287,31 +287,30 @@ class Menu extends hx.EventEmitter
           checkEvent(e, self)
         self.emit 'keydown', e
 
-      self.dropdown._.dropdown
-        .on 'click', 'hx.menu', (e) ->
-          # get the closest menu item - uses nodes as blank selection can be
-          # returned if the target is a hx-menu-item
+      self.dropdown._.dropdown?.on 'click', 'hx.menu', (e) ->
+        # get the closest menu item - uses nodes as blank selection can be
+        # returned if the target is a hx-menu-item
 
-          target = if hx.select(e.target).classed('hx-menu-link')
-            e.target
-          else
-            hx.select(e.target).closest('.hx-menu-item').node()
+        target = if hx.select(e.target).classed('hx-menu-link')
+          e.target
+        else
+          hx.select(e.target).closest('.hx-menu-item').node()
 
-          if target
+        if target
 
-            index = -1
+          index = -1
 
-            t = hx.select(target)
-            if t.classed('hx-menu-link')
-              allItems = getAllItems(self)
-              i = 0
-              while index < 0 and i < allItems.length
-                if allItems[i].node is target
-                  index = i
-                  break
-                i += 1
-              setActive(self, index, undefined, true)
-            targetElem.node().focus()
+          t = hx.select(target)
+          if t.classed('hx-menu-link')
+            allItems = getAllItems(self)
+            i = 0
+            while index < 0 and i < allItems.length
+              if allItems[i].node is target
+                index = i
+                break
+              i += 1
+            setActive(self, index, undefined, true)
+          targetElem.node().focus()
 
 
     # pipe the dropdown events through
--- a/modules/view/main/index.coffee
+++ b/modules/view/main/index.coffee
@@ -71,23 +71,22 @@ class View
       viewEnterWarning = (element, selector) ->
         hx.consoleWarning "view enter fn returned", element, "! It didn't match selector", selector, ", so you may encounter odd behavior"
 
+      classes = @selector.split('.')
+      selectorContainsClasses = classes.length > 1
+      classString = classes.slice(1).join(' ')
+
       newNodeSet = enterSet.map (d, i) =>
         datum = d.datum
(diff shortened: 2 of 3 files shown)


modules/menu/main/index.coffee
```diff
@@ -287,31 +287,30 @@ class Menu extends hx.EventEmitter
...
-              allItems = getAllItems(self)
-              i = 0
-              while index < 0 and i < allItems.length
-                if allItems[i].node is target
-                  index = i
-                  break
-                i += 1
-              setActive(self, index, undefined, true)
-            targetElem.node().focus()
+      self.dropdown._.dropdown?.on 'click', 'hx.menu', (e) ->
```
c-frater: I added an existence check here as I noticed a few errors when trying larger autocompletes

modules/view/main/index.coffee
```diff
@@ -2,12 +2,12 @@
 class View
   constructor: (@rootSelection, @selector, @defaultType) ->
     self = this
+    elementType = self.selector.split('.')[0]
+    classes = self.selector.split('.').slice(1).join ' '
```
c-frater: These don't change once you've made the view so we split the selector outside the `@new` function
This also removes the de-structuring and improves performance across larger numbers of nodes.

modules/view/main/index.coffee
```diff
@@ -68,14 +68,26 @@ class View
...
-          hx.consoleWarning "view enter fn returned", element, "! It didn't match selector", @selector, ", so you may encounter odd behavior"
+        if @selector.indexOf('.') >= 0
+          classes = @selector.split('.')
+          isChild = @rootSelection.node().contains(element)
+
+          # Checks isChild first as it's the quickest operation
+          if not isChild
+            viewEnterWarning(element, @selector)
+          else
+            isClassedCorrectly = hx.select(element).classed(classes.slice(1).join(' '))
```
c-frater: The first item in the array is either an empty string or the element type of the element (i.e. for `div.class` it would be `div` whereas for `.class` it would be empty string). We also care about all the classes in the selector so we check them all here using `.classed`

modules/view/main/index.coffee
```diff
@@ -68,14 +68,26 @@ class View
...
+            if typeof isClassedCorrectly isnt 'boolean'
+              isClassedCorrectly = isClassedCorrectly.every((e) -> e)
```
c-frater: If there is an array of classes, `isClassedCorrectly` is initially an array. In the name of speed, we check if it's not a boolean rather than checking if it is an array.

modules/view/test/spec.coffee
```diff
@@ -1,19 +1,50 @@
...
+    hx.consoleWarning = chai.spy()
+
+  afterEach ->
+    hx.consoleWarning = origConsoleWarning
+
   it "should create elements of a given type using the default view enter fn", ->
     selection = hx.detached('div')
     selection.view('element-type').apply([1..3])
     selection.selectAll('element-type').size().should.equal(3)
+    hx.consoleWarning.should.not.have.been.called()
```
c-frater: Added this to every function because it shouldn't be called in these situations. We are also now testing for cases when it _should_ be called.

modules/menu/main/index.coffee
jmsmyth: Any idea what would cause _dropdown to not be there?

c-frater: Pretty sure it was because I'd added an autocomplete to a button so it wasn't passing an autocomplete check and hiding the dropdown before this was being called, pretty sure it was caused by my poor use of autocomplete on a button rather than this code but I figured it should probably have the check as all the other uses check for its existence as well :)

modules/view/main/index.coffee
jmsmyth: `classes = @selector.split('.')` and `classes.slice(1).join(' ')` could be moved outside of the loop

modules/view/main/index.coffee
```diff
@@ -68,14 +68,26 @@ class View
...
+        hx.consoleWarning "view enter fn returned", element, "! It didn't match selector", selector, ", so you may encounter odd behavior"
 
       newNodeSet = enterSet.map (d, i) =>
         datum = d.datum
         element = @new.call @rootSelection, d.datum, i
-        subselection = @rootSelection.selectAll @selector
 
-        unless element in subselection.nodes
-          hx.consoleWarning "view enter fn returned", element, "! It didn't match selector", @selector, ", so you may encounter odd behavior"
+        if @selector.indexOf('.') >= 0
```
jmsmyth: the `indexOf` check could be moved outside of the loop

modules/view/main/index.coffee
jmsmyth: Shouldn't `isClassedCorrectly` always be a boolean?

c-frater: Yes, I originally had it checking an array and for some reason had it in my head that it was still returning an array... I've changed this :)

jmsmyth:
LGTM

PR #158, opened 2016-03-07
Fix the min/max options for number-picker + add some tests

jmsmyth:


modules/number-picker/main/index.coffee
```diff
@@ -2,8 +2,8 @@ checkValue = (numberPicker, context) ->
   value = oldValue = context.value()
   max = numberPicker.max()
   min = numberPicker.min()
-  if max isnt undefined then value = hx.min([value, max])
-  if min isnt undefined then value = hx.max([value, min])
+  if max isnt undefined then value = Math.min(value, max)
+  if min isnt undefined then value = Math.max(value, min)
```
georgesimms: Can we use hx.clamp here?

Also `if max?` or `if hx.defined(max)`

c-frater: As far as I can remember, it doesn't use `hx.clamp` because you may not always have both a min and a max at the same time which is required by `hx.clamp`. 
Also: 
```coffeescript
if max?
 a
if max isnt undefined
 b
```
evaluates to
```js
if (typeof max !== "undefined" && max !== null) {
  a;
}

if (max !== void 0) {
  b;
}
```
In this situation, `max` has a default value of undefined so, while it is a bit nicer to read, it's more efficient to use `max isnt undefined`
The difference is negligible and I don't really mind either way, it's just worth noting that coffeescript does this a lot - makes the code easier to read but adds complexity to otherwise simple functions.

PR #164, opened 2016-03-10
Ensure that the locale is initialised to something

jmsmyth:
When navigator.language is set to a non supported locale, if will fall back to 'en'

### DIFF
--- a/modules/preferences/main/index.coffee
+++ b/modules/preferences/main/index.coffee
@@ -207,8 +207,8 @@ class Preferences extends hx.EventEmitter
       modal: modal
     }
 
-    defaultLocaleId = moment?.locale() or navigator.language or 'en'
-    if not lookupLocale(defaultLocaleId)
+    defaultLocaleId = moment?.locale() or navigator.language
+    if not (hx.isString(defaultLocaleId) and lookupLocale(defaultLocaleId))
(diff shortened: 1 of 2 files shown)


modules/preferences/main/index.coffee
```diff
@@ -207,7 +207,10 @@ class Preferences extends hx.EventEmitter
       modal: modal
     }
 
-    @locale moment?.locale() or navigator.language or 'en'
+    defaultLocaleId = moment?.locale() or navigator.language or 'en'
```
c-frater: probably doesn't need the `or 'en'` on this line as navigator.language will (afaik) always be defined

c-frater:
Looks fine, the only comment is more of a slight tweak than an issue :)

modules/preferences/main/index.coffee
jmsmyth: Although it should be always defined, just to be safe it might be best to leave it in as there is no knowing what phantom and slimer will return for navigator.language

c-frater: The line below where it checks 'localeLookup' could just check `defaultLocale` first to save _always_ looking up the locale when we know 'en' is in the list

jmsmyth: Good thinking - will update

Issue #142, opened 2016-02-16
Modal: calling modal.show() multiple times causes the modal content to duplicate


c-frater:
This is evident mostly when clicking a button to show a modal then pressing 'enter' (while the button has focus) opens multiple modals.

This should be fixed by:
- clearing focus when a modal is shown
- clearing existing modals when a modal is shown (or checking if it's already open and not doing anything)

To reproduce, go to https://www.hexagonjs.io/docs/1.1.0/modal/, click 'show modal' then press 'enter' or call `modal.show()` multiple times

c-frater:
Pushed in 1.2.1 as it was ready when the release was made

PR #165, opened 2016-03-11
Modal: prevent content from being duplicated

c-frater:
Fix for #142

### DIFF
--- a/modules/modal/main/index.coffee
+++ b/modules/modal/main/index.coffee
@@ -24,21 +24,19 @@ class Modal extends hx.EventEmitter
 
 
   show: (cb) ->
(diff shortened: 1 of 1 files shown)


jmsmyth:
Looks good

modules/modal/main/index.coffee
```diff
@@ -24,21 +24,19 @@ class Modal extends hx.EventEmitter
 
 
   show: (cb) ->
+    # Clears the focus so pressing 'enter' does not cause buttons to call modal.show()
+    document.activeElement.blur()
```
georgesimms: Can document.activeElement be null, eg if the modal was shown programmatically ?

Issue #25, opened 2015-12-08
New Component: File Input


c-frater:
The `<input type="file" />` not only looks pretty awful but could do with some extra functionality, like drag and drop support. 

This should be added as a standalone module as well as a component that can be added to Form Builder.

PR #166, opened 2016-03-11
New Component: File Input

c-frater:
As per #25


assets/hexagon-icons(icomoon).json
```diff
@@ -4,6 +4,32 @@
 		{
```
c-frater: Added the 'upload' icon to the hexagon icons fonts

modules/file-input/main/index.coffee
```diff
@@ -0,0 +1,165 @@
...
+    if resolvedOptions.multiple and not options.buttonText
+      resolvedOptions.buttonText = 'Choose Files'
+
+    selection = hx.select @selector
+      .classed 'hx-file-input', true
+
+    input = hx.detached 'input'
+      .class 'hx-file-input-hidden'
+      .attr 'type', 'file'
+      .attr 'accept', resolvedOptions.accept
```
c-frater: We add this but there doesn't seem to be a consistent way of checking validity? (this is an issue when dragging/dropping files as we can't validate)

modules/file-input/main/index.coffee
```diff
@@ -0,0 +1,165 @@
...
+        e.preventDefault()
+        e.stopPropagation()
+
+      drop = (e) =>
+        preventDefault(e)
+        unless @_.disabled
+          files = e.dataTransfer.files
+          validates = true
+          for file in files
+            break if not validates
```
c-frater: This validation loop currently does nothing - I'm investigating how to do this whilst retaining the current 'accepts' property of the file input

modules/file-input/main/index.coffee
```diff
@@ -0,0 +1,165 @@
...
+        .add(button)
+        .add(selectedFiles))
+      .add(dropdownDiv)
+
+  disabled: (disabled) ->
+    if arguments.length
+      disabledAttr = if disabled then 'disabled' else undefined
+      @_.input.attr 'disabled', disabledAttr
+      @_.button.attr 'disabled', disabledAttr
+      @_.disabled = disabled
```
c-frater: Need to add a class here to make the 'no files' text greyed out
Could also remove the file (if set) but not sure that's necessary?

modules/file-input/main/index.coffee
```diff
@@ -0,0 +1,165 @@
...
+      this
+    else
+      @_.disabled or false
+
+  value: -> @_.fileMap.values()
+
+
```
c-frater: Could do with a 'clear' or 'reset' method for programatically removing the value from the field.

modules/file-input/main/index.coffee
```diff
@@ -0,0 +1,165 @@
...
+    if arguments.length
+      disabledAttr = if disabled then 'disabled' else undefined
+      @_.input.attr 'disabled', disabledAttr
+      @_.button.attr 'disabled', disabledAttr
+      @_.disabled = disabled
+      this
+    else
+      @_.disabled or false
+
+  value: -> @_.fileMap.values()
```
c-frater: As per security requirements, it is not possible to set the value of a file input so we have not exposed that functionality.

modules/file-input/module.json
```diff
@@ -0,0 +1,11 @@
+{
+  "dependencies": [
+    "component",
+    "selection",
+    "event-emitter"
```
c-frater: Need to add dependencies here for Map/Util (and any others that were missed)

c-frater:
This is currently a bit WIP but I thought I'd push it to get feedback on things I can change whilst polishing it off.
I've added comments where there are things to do, I'm sure there are things I've missed/could do better so feel free to comment/criticise :)

modules/file-input/main/index.coffee
georgesimms: What will happen if someone tries ? Console warning?

georgesimms:
Screenshot?

georgesimms:
Also, this may be a more general issue, but there is copypasta  in the disabled methods - is this avoidable ?

modules/file-input/main/index.coffee
c-frater: It returns the value of the file input as if they hadn't passed in a value. 
We could add a console warning but I didn't see the point warning the user that they've tried to do something that isn't documented.

c-frater:
From top to bottom:
- Default file input, nothing entered
- 'multiple' file input, nothing entered
- Default file input with file added
- 'multiple' file input with values entered and dropdown open (dropdown is only shown when there is more than 1 file)

file-input

modules/file-input/test/spec.coffee
```diff
@@ -0,0 +1 @@
+describe "File Input", ->
```
jmsmyth: Could do with tests once you're happy with how it works

c-frater: Yeah, I plan to add some - although unfortunately there's no way to 'set' the files to use due to security issues so it's going to be fun to test everything...

c-frater:
@jmsmyth Pretty sure this is done now I've added it to form builder, is there anything you can see that still needs doing?

PR #167, opened 2016-03-14
Update package.json engine range

jmsmyth:
Change the supported engine range in package.json to support all 4.x.x versions


c-frater:
LGTM :)

PR #168, opened 2016-03-14
Update list of node versions to test in travis

jmsmyth:
It will now test only the latest version of 4.x.x (along with 0.12,  0.11). iojs support dropped


c-frater:
Do we need to test 0.11 still if the package.json engines says `>=0.12`?

jmsmyth:
Probably not - lets remove it as well

PR #169, opened 2016-03-14
Enable building against version 5 of node in travis

jmsmyth:
I expect this to fail initially


jmsmyth:
Well, that went better than expected :-p

Issue #196, opened 2016-04-07
Palette: add borderContext function


c-frater:
A function exists for background and text context but there isn't an equivalent for border, despite palette including border context classes.

Issue #193, opened 2016-04-01
Tabs: Divs with hx-tabs-content lose their other classes when the page loads


idanov:
It seems that the javascript setting the classes of a div marked as `hx-tabs-content` ignores all other classes except the tab related.

@jmsmyth  pointed to this line: https://github.com/ocadotechnology/hexagonjs/blob/master/modules/tabs/main/index.coffee#L8

Issue #184, opened 2016-04-01
Slider: emits slidestart events on change


jmsmyth:
`slidestart` should only be emitted when the sliding starts, but it appears to emitted with every `change` event

Issue #183, opened 2016-04-01
Slider: emits events even when there is no change


jmsmyth:
To reproduce: drag the slider past the end of the slider - events will be emitted when the mouse is moved, even though the slider value doesn't change

PR #197, opened 2016-04-07
Bugfixes

c-frater:
Bugfixes for a selection of issues. Also includes the addition of `hx.palette.borderContext`.

#133, #146, #161, #162, #178, #182, #183, #184, #193, #196, #202


modules/form-builder/main/index.coffee
```diff
@@ -325,9 +325,7 @@ class Form extends hx.EventEmitter
         switch it.type
           when 'checkbox' then hx.select(node).prop('checked', value)
           when 'radio' then hx.select(node).selectAll('input').filter((d) -> d.value() is value).prop('checked', true)
-          when 'tagInput'
-            tagInput = hx.component(it.extras.componentNode or node)
-            value.forEach((e) -> tagInput.add(e))
+          when 'tagInput' then hx.component(it.extras.componentNode or node).items(value)
```
c-frater: Set the tags to the value instead of adding a new tag for each value.

modules/pivot-table/main/index.coffee
```diff
@@ -43,26 +22,32 @@ class PivotTable extends hx.EventEmitter
     @tableHead = @table.append('thead')
     @tableBody = @table.append('tbody')
 
+    if @options.data?
+      @data(@options.data)
```
c-frater: Allow setting of data in the options

modules/pivot-table/main/index.coffee
```diff
@@ -43,26 +22,32 @@ class PivotTable extends hx.EventEmitter
...
 
       @_.data = data
 
-      topData = if data.topHead?.length > 0 then data.topHead else []
-      leftData = if data.leftHead?.length > 0 then data.leftHead else []
+      clonedData = hx.merge(data, true)
+
+      topData = if clonedData.topHead?.length > 0 then clonedData.topHead else []
+      leftData = if clonedData.leftHead?.length > 0 then clonedData.leftHead else []
+      bodyData = if clonedData.body?.length > 0 then clonedData.body else []
```
c-frater: Prevent data mutation by using cloned data

modules/pivot-table/main/index.coffee
```diff
@@ -1,24 +1,52 @@
...
+
+  headView = head.view('tr')
+    .enter (data) -> rowViewEnter.call(this, data, true)
+    .update (data, node) -> rowViewUpdate(data, node, true)
+
+  bodyView = body.view('tr')
+    .enter (data) -> rowViewEnter.call(this, data, false)
+    .update (data, node) -> rowViewUpdate(data, node, false)
+
+  [headView, bodyView]
```
c-frater: Made this a function that returns views to reduce clutter in the data method and reduce the amount of variables that need to be passed down the chain to the cells (namely cellRenderer and leftShifted)

modules/pivot-table/main/index.coffee
```diff
@@ -1,24 +1,51 @@
...
```
c-frater: Created a function to return the views so that they can be re-used rather than having to create new ones every time the data is set.

modules/pivot-table/main/index.coffee
```diff
@@ -1,24 +1,51 @@
...
+      selection.text('')
+    else table.options.cellRender(data, node, cellIsHead, index) # Head / Body cells
+
+  rowViewEnter = (data, isHead) ->
+    rowNode = @append('tr').node()
+    rowView = hx.select(rowNode).view('.hx-pivot-table-cell')
+      .enter (datum, index) ->
+        # index starts at 0 for new cells added to a row so we override the
+        # value here.
+        index = data.indexOf(datum)
```
c-frater: Not sure if this is a bug with views or intentional behaviour? @jmsmyth

modules/pivot-table/main/index.coffee
```diff
@@ -43,42 +71,44 @@ class PivotTable extends hx.EventEmitter
     @tableHead = @table.append('thead')
     @tableBody = @table.append('tbody')
 
+    # Create the re-usable views
+    [@tableHeadView, @tableBodyView] =
+      createTableView(this, @tableHead, @tableBody)
+
+    if @options.data?
+      @data(@options.data)
```
c-frater: Allow setting of data in the options, can be done in the standard and fluid versions

modules/pivot-table/main/index.coffee
```diff
@@ -43,42 +71,44 @@ class PivotTable extends hx.EventEmitter
...
+      _.leftShifted = false
 
-      @_.data = data
+      clonedData = hx.merge(data, true)
 
-      topData = if data.topHead?.length > 0 then data.topHead else []
-      leftData = if data.leftHead?.length > 0 then data.leftHead else []
+      topData = if clonedData.topHead?.length > 0 then clonedData.topHead else []
+      leftData = if clonedData.leftHead?.length > 0 then clonedData.leftHead else []
+      bodyData = if clonedData.body?.length > 0 then clonedData.body else []
```
c-frater: Use cloned data to prevent data mutation

modules/preferences/main/index.coffee
```diff
@@ -211,7 +211,15 @@ class Preferences extends hx.EventEmitter
...
-    @timezone moment?.tz?.guess() or 'UTC+00:00'
+
+    guessedMomentTimezone = moment?.tz?.guess()
+    if guessedMomentTimezone?
+      @supportedTimezones moment.tz.names()
+      @timezoneOffsetLookup (timezone, datestamp) ->
+        -(moment.tz.zone(timezone).offset(timestamp) / 60)
+      @timezone guessedMomentTimezone
+    else
+      @timezone 'UTC+00:00'
```
c-frater: Correctly guess the timezone using moment and update the supported timezones to use the list from moment.tz when moment is present

modules/pivot-table/main/index.coffee
georgesimms: can we use hx.clone here ?

modules/pivot-table/main/index.coffee
georgesimms: Add a test that documents the behavior, at least, so if it changes you know

modules/pivot-table/main/index.coffee
georgesimms: Also, add a regression test for the data mutating

modules/slider/main/index.coffee
```diff
@@ -62,9 +60,10 @@ slide = (e) ->
     _.dragging = 'value'
     val.value = posToValue.call this, pos
 
+  prevValue = @value()
   @value(val)
-
-  @emit 'change', @value()
+  if prevValue isnt @value()
+    @emit 'change', @value()
   return
```
georgesimms: superfluous return again ?

modules/palette/test/spec.coffee
```diff
@@ -90,4 +90,35 @@ describe 'palette', ->
       selection = hx.detached('div').class('hx-background-positive hx-background-negative')
       hx.palette.backgroundContext(selection, undefined)
       selection.classed('hx-background-positive').should.equal(false)
-      selection.classed('hx-background-negative').should.equal(false)
\ No newline at end of file
+      selection.classed('hx-background-negative').should.equal(false)
+
+  describe 'hx.palette.borderContext should correctly class an element', ->
```
georgesimms: +1 for adding tests...

modules/palette/main/index.coffee
```diff
@@ -34,4 +34,14 @@ hx.palette.backgroundContext =  (selector, context) ->
   else
     for c in paletteContexts
       if selection.classed('hx-background-' + c) then return c
+    return undefined
+
+hx.palette.borderContext =  (selector, context) ->
+  selection = flatSelect(selector)
+  if arguments.length > 1
+    paletteContexts.forEach (c) -> selection.classed('hx-border-' + c, c is context)
```
georgesimms: or:
```
paletteContextsByRelevance = new hx.Map hx.groupBy paletteContexts, (c) -> c is context
hx.consoleWarning "#{context} is not a known context!" unless relevantPaletteContexts.get(true).length
paletteContextsByRelevance.entries().forEach (desiredState, contexts) ->
  contexts.forEach (context) ->
    selection.classed "hx-border-#{context}", desiredState
```

demo/plain/index.js
```diff
@@ -1,30 +0,0 @@
-container = hx.select('body').append('div')
```
georgesimms: What is going on here ?

c-frater: Some leftover code in the 'Sandbox' has been removed

modules/pivot-table/main/index.coffee
c-frater: Completely forgot about hx.clone... I'll use that instead :)

modules/slider/main/index.coffee
c-frater: I guess so - although it's nice to explicitly return nothing rather than letting coffeescript return _something_, especially as I don't know if events use / change behaviour based on the return value

modules/palette/test/spec.coffee
c-frater: Pretty much copy pasta of the backgroundContext tests ;)

modules/palette/main/index.coffee
c-frater: Could actually get away with:
```coffeescript
mappedContexts = palleteContexts.map (c) -> "hx-border-#{c}"
selection.classed(mappedContext, false)
selection.classed("hx-border-#{context}", true) if context in paletteContexts
```

c-frater:
Pretty sure I've done everything mentioned in the comment(s)
- Add test to view for expected behaviour (and tweak the pivot table view)
- Use clone instead of merge
- Use better code for the palette functions and add the warning (with tests)

modules/slider/main/index.coffee
georgesimms: `undefined` might be a better way to achieve this.

c-frater: This is why that's not the case:
```coffeescript
a = (a) ->
  return

a = (a) ->
  undefined
```
```js
var a;

a = function(a) {};

a = function(a) {
  return void 0;
};
```

Having an empty return statement is the correct way of returning nothing in coffeescript.

PR #198, opened 2016-04-07
Selection: add shallowSelect and shallowSelectAll

c-frater:
#104

### DIFF
--- a/modules/selection/main/index.coffee
+++ b/modules/selection/main/index.coffee
@@ -16,22 +16,15 @@ getMethod = (node, methodName) ->
   else
     Element.prototype[methodName]
 
-getMatches = (node) ->
-  # XXX: use hx.vendor
-  matchesMethod =
-    if node.matches then 'matches'
-    else if node.matchesSelector then 'matchesSelector'
-    else if node.mozMatchesSelector then 'mozMatchesSelector'
-    else if node.webkitMatchesSelector then 'webkitMatchesSelector'
-    else if node.oMatchesSelector then 'oMatchesSelector'
-    else if node.msMatchesSelector then 'msMatchesSelector'
 
-  if matchesMethod then getMethod(node, matchesMethod)
-  else (selector) ->
-    matchingNodes = getMethod(this, 'querySelector').call(this, selector)
-    i = matchingNodes.length
-    i while --i >= 0 and matchingNodes.item(i) isnt this
-    i > -1
+# Should only be called with Function.call(node, selector)
+matchPolyfill = (selector) ->
+  node = this
+  matchingNodes = (node.document or node.ownerDocument).querySelectorAll(selector)
+  [].slice.call(matchingNodes).indexOf(node) > -1
+
+getMatches = (node) ->
+  node.matches or hx.vendor(node, 'matchesSelector') or matchPolyfill
 
 selectSingle = (selector, node) -> getMethod(node, 'querySelector').call(node, selector)
 selectAll = (selector, node) -> getMethod(node, 'querySelectorAll').call(node, selector)
--- a/modules/selection/main/index.coffee
+++ b/modules/selection/main/index.coffee
@@ -16,8 +16,40 @@ getMethod = (node, methodName) ->
   else
     Element.prototype[methodName]
 
+getMatches = (node) ->
+  # XXX: use hx.vendor
+  matchesMethod =
+    if node.matches then 'matches'
+    else if node.matchesSelector then 'matchesSelector'
+    else if node.mozMatchesSelector then 'mozMatchesSelector'
+    else if node.webkitMatchesSelector then 'webkitMatchesSelector'
+    else if node.oMatchesSelector then 'oMatchesSelector'
+    else if node.msMatchesSelector then 'msMatchesSelector'
+
+  if matchesMethod then getMethod(node, matchesMethod)
+  else (selector) ->
+    matchingNodes = getMethod(this, 'querySelector').call(this, selector)
+    i = matchingNodes.length
+    i while --i >= 0 and matchingNodes.item(i) isnt this
+    i > -1
+
 selectSingle = (selector, node) -> getMethod(node, 'querySelector').call(node, selector)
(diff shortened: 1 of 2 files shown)


modules/selection/main/index.coffee
```diff
@@ -16,9 +16,41 @@ getMethod = (node, methodName) ->
...
 
+getMatches = (node) ->
+  # XXX: use hx.vendor
+  matchesMethod =
+    if node.matches then 'matches'
+    else if node.matchesSelector then 'matchesSelector'
+    else if node.mozMatchesSelector then 'mozMatchesSelector'
+    else if node.webkitMatchesSelector then 'webkitMatchesSelector'
+    else if node.oMatchesSelector then 'oMatchesSelector'
+    else if node.msMatchesSelector then 'msMatchesSelector'
```
georgesimms: can we use hx.vendor here ?

modules/selection/main/index.coffee
```diff
@@ -16,9 +16,41 @@ getMethod = (node, methodName) ->
...
+
+  if matchesMethod then getMethod(node, matchesMethod)
+  else (selector) ->
+    matchingNodes = getMethod(this, 'querySelector').call(this, selector)
+    i = matchingNodes.length
+    i while --i >= 0 and matchingNodes.item(i) isnt this
+    i > -1
```
georgesimms: The indentation of this if block is a bit hard to follow, maybe

```
if matchesMethod
  getMethod node, matchesMethod
else
  (selector) ->
    #stuff

File: modules/selection/main/index.coffee [hunk 3]
```diff
@@ -16,9 +16,41 @@ getMethod = (node, methodName) ->
...
+    else if node.mozMatchesSelector then 'mozMatchesSelector'
+    else if node.webkitMatchesSelector then 'webkitMatchesSelector'
+    else if node.oMatchesSelector then 'oMatchesSelector'
+    else if node.msMatchesSelector then 'msMatchesSelector'
+
+  if matchesMethod then getMethod(node, matchesMethod)
+  else (selector) ->
+    matchingNodes = getMethod(this, 'querySelector').call(this, selector)
+    i = matchingNodes.length
+    i while --i >= 0 and matchingNodes.item(i) isnt this
```
georgesimms: * What is the type of matchingNodes here ? is it a pseudo-array ?
* Why do we start at the end and work backward ?
* Can we
```
this in Array::slice.call matchingNodes
```
?

modules/selection/main/index.coffee
```diff
@@ -16,9 +16,41 @@ getMethod = (node, methodName) ->
...
+  for child in node.children
+    if matchFn.call(child, selector)
+      return child
+
+shallowSelectAll = (selector, node) ->
+  matchFn = getMatches(node)
+  matchingNodes = []
+  for child in node.children
+    if matchFn.call(child, selector)
+      matchingNodes.push child
```
georgesimms: Again, maybe
```
Array::slice.call node.children
  .filter (selector) -> matchFn.call child, selector
```

modules/selection/main/index.coffee
```diff
@@ -16,9 +16,41 @@ getMethod = (node, methodName) ->
...
+    if node.matches then 'matches'
+    else if node.matchesSelector then 'matchesSelector'
+    else if node.mozMatchesSelector then 'mozMatchesSelector'
+    else if node.webkitMatchesSelector then 'webkitMatchesSelector'
+    else if node.oMatchesSelector then 'oMatchesSelector'
+    else if node.msMatchesSelector then 'msMatchesSelector'
+
+  if matchesMethod then getMethod(node, matchesMethod)
+  else (selector) ->
+    matchingNodes = getMethod(this, 'querySelector').call(this, selector)
```
jmsmyth: `this` isn't the right thing to be passing in here...

modules/selection/main/index.coffee
jmsmyth: Using `.filter` will probably be less performant

modules/selection/main/index.coffee
c-frater: Potentially, although it looks like it would have to still check for 'node.matches' first and then use hx.vendor if that doesn't exist

modules/selection/main/index.coffee
c-frater: +1. It might also be worth double checking that this function works in the browser(s) that don't support the 'real' matches (If there are any left...)

modules/selection/main/index.coffee
c-frater: The 'polyfill' was taken from the MDN which is why it's a bit odd:
```js
​function matches(elm, selector) {
    var matches = (elm.document || elm.ownerDocument).querySelectorAll(selector),
        i = matches.length;
    while (--i >= 0 && matches.item(i) !== elm) {}
    return i > -1;
}
```

Looking at it, the implementation is doing something different from the polyfill anyway (which is wrong and won't work)

modules/selection/main/index.coffee
c-frater: +1 - I wanted to keep it efficient by using a for loop :)

modules/selection/main/index.coffee
c-frater: Yeah, this 'polyfill' is wrong for multiple reasons....

modules/selection/main/index.coffee
```diff
@@ -16,9 +16,34 @@ getMethod = (node, methodName) ->
...
 
+
+# Should only be called with Function.call(node, selector)
+matchPolyfill = (selector) ->
+  node = this
+  matchingNodes = (node.document or node.ownerDocument).querySelectorAll(selector)
+  [].slice.call(matchingNodes).indexOf(node) > -1
+
+getMatches = (node) ->
+  node.matches or hx.vendor(node, 'matchesSelector') or matchPolyfill
```
c-frater: Uses hx.vendor - This is assuming that we don't need to do the 'getMethod' style of call?

Issue #180, opened 2016-03-23
Dropdown: improve reusability of dropdowns


c-frater:
Current dropdowns are difficult to reuse reliably. It would be nice to have the ability to change the setup function and manually call render 'on the fly'
This would involve adding:
```js
hx.Dropdown::render([callback: Function]) // Render the dropdown using the current renderer (without closing/opening)
hx.Dropdown::setupDropdown(content) // Set the dropdown content
hx.Dropdown::setupDropdown() // Get the current dropdown content
```
This would need to deal with the fact that the dropdown content can be a string or a function.
Also, setting the 'setupDropdown' function whilst the dropdown is open should also potentially call render

georgesimms:
Would the idea here to be to render in a detached div, and then just add it to the dom when show is called ?

c-frater:
This issue is mainly for re-rendering the dropdown when it's already open and less about reusing the actual selection. 
It might cause a lot of overhead to render dropdowns before they are needed (e.g. if you have 500 dropdowns on the page, you don't want to store all the rendered content in the detached dom if you're not going to open them)

georgesimms:
Right, so render shouldn't do anything if the dropdown isn't open yet ?

c-frater:
The use case for this is in `Autocomplete Picker`:

We want to be able to re-render the dropdown without having to do this:

```coffeescript
renderMenu = (items) ->
        menu.items(items)
        menu.dropdown._.setupDropdown(menu.dropdown._.dropdown.node())
```

georgesimms:
Right

Issue #26, opened 2015-12-08
All: Make all asynchronous content use Promises 


c-frater:
Animations are probably the main candidate but there are other things that could do with being Promisified.

jmsmyth:
I think this change has enough api breaking potential to be in the 2.0.0 milestone.

georgesimms:
Can this be split into two sections, morphs and everything else? The fact that morphs are not promises is sometimes confusing and can lead to subtle bugs.

jmsmyth:
Sounds like a plan. I think the whole animation module needs a revamp as there are clearly a few issues with it at the moment.

One possible approach is deprecating the entire animate module, and adding a couple more overloads to the selection api:

    hx.select(node).style(property, value, duration, [easeFunc]): Promise
    hx.select(node).prop(property, value, duration, [easeFunc]): Promise
    hx.select(node).attr(property, value, duration, [easeFunc]): Promise

or 

    hx.select(node).style(property, value, {
      duration: <duration-in-ms>,
      ease: <easeFunc>
    }): Promise

Then the chaining of animations can happen via combinations of `Promise.all`, `Promise.race` and `then`. I'm pretty sure this approach is not api breaking so could theoretically go into some 1.x.0 release. 

The only thing isn't clear how to cancel animations with this approach. Not sure how that can be exposed. Any thoughts?

c-frater:
I'm not sure if standard Promises are cancellable by default but we could always overload them and add a `cancel` method?

Also, I prefer the 2nd method (using an options object) as it would allow adding of any future options if required without needing any param changes :)

PR #199, opened 2016-04-08
Autocomplete picker

### LINKED ISSUE #17: New Component: Picker/Autocomplete hybrid
A combination of the picker and autocomplete functionality for selecting a value. This would enable the user to search for values in a dropdown to apply a value from very large lists.

c-frater:
Resolves #17 
Adds an input field inside the picker dropdown so that the user can filter the values in the picker.

UI:
autocomplete-picker

UX:
- Input is focused when picker is opened
- User can press up/down to select menu items
- Typing filters menu items
- When user types, if there are values then the 'active item' is set to the first item so pressing enter will select it as the picker value
- When there is no value in the input, pressing enter will not trigger the 'select value' unless the user has manually highlighted an item with the arrow keys
- If there are no values, 'No Results Found' (configurable) text is shown:

autocomplete-picker-filter


georgesimms:
+1

really positive, I like it.

* Do you have an overview of what the API looks like ?
* Is it possible for the legal values to be fetched asynchronously for a given input, eg over AJAX ?

c-frater:
It works in exactly the same way as the picker currently.
I'm still not sure it's the best implementation and might instead add 'picker' functionality to the autocomplete in a similar way - showing the input field in the dropdown and having the 'value' as a picker.

c-frater:
Ok so there are two ways to go about doing this, it depends on the use cases

#### Add 'picker' functionality to autocomplete
##### Pros
- Retain all autocomplete functionality (eventual matching etc)
- Keeps all autocomplete functionality in one module

##### Cons
- Potentially complex to implement
- Has to use 'must match'
- Doesn't keep to the `<select>` methodology (users may not be expecting it to work that way) 
- Complex to set up for 'static' lists

#### Add 'autocomplete' functionality to picker
##### Pros
- Easier to implement
- Retains 'async' items functionality (probably...)
- Simpler implementation - only have to pass in one option when using default structure for items

##### Cons
- Loses full autocomplete functionality as dropdown inside a dropdown doesn't work nicely for UX
- Potentially too simple of an implementation (doesn't give a lot of flexibility, can't return values based on the search term)

jmsmyth:
There is also the third option of implementing as a new module..

#### New Module

##### Pros
 - Doesn't add any complexity to the existing modules
 - Easier to test (adding to the other modules will increase the number of combinations of options to test) 
 - Not tied to the behaviour of the other module

##### Cons
 - More work (maybe? although in the long run it might not be due to having smaller simpler modules)
 - Autocomplete code might have to be duplicated / a way to share autocomplete code would need figuring out.

Async/getting autocomplete values from the server is definitely needed - the request for this module was first raised for this reason

c-frater:
Thinking more about this, both the autocomplete and the picker have async functionality currently - albeit a bit simpler for the picker. 
I think it probably does make the most sense to make a new module, I'll revert the picker changes and create a new module that looks/behaves as described in here. I'll look into ways to prevent duplication in the autocomplete code - It looks like it may be done the same way as `date-localizer` (i.e. give it it's own module) and then work out a better way to share functionality once we've worked more on the ES6 side of things.

c-frater:
Reverted out of the changes made to picker but left in the update to the internals (as I still think it makes sense for most variables to be stored on `_` instead of the picker itself)

modules/autocomplete-picker/main/index.coffee
```diff
@@ -0,0 +1,210 @@
...
+        unselectable: true
+
+      otherResultsItem =
+        text: resolvedOptions.otherResultsText
+        unselectable: true
+        heading: true
+
+      renderMenu = (items) ->
+        menu.items(items)
+        menu.dropdown._.setupDropdown(menu.dropdown._.dropdown.node())
```
c-frater: This function will not be required once #180 has been completed as `menu.items(items)` should re-render the dropdown if it is called when the dropdown is open.

modules/autocomplete-picker/main/index.coffee
```diff
@@ -0,0 +1,210 @@
...
+
+      populateMenu = (term) ->
+        renderMenu([loadingItem])
+        feed.filter term, (results, otherResults) ->
+          if results.length is 0
+            results.push(noResultsItem)
+          if otherResults.length > 0
+            otherResults = [otherResultsItem].concat(otherResults)
```
c-frater: The feed decides whether otherResults should be returned

modules/autocomplete-feed/module.json
```diff
@@ -0,0 +1,14 @@
+{
+  "dependencies": [
+    "filter",
+    "map",
+    "sort",
+    "util"
+  ],
+  "keywords": [
```
jmsmyth: fyi, keywords aren't actually needed here (despite every module having them). The ones used by the docs are defined elsewhere

modules/autocomplete-feed/main/index.coffee
```diff
@@ -0,0 +1,99 @@
...
+    resolvedOptions.filter ?= (items, term) ->
+      hx.filter[resolvedOptions.matchType](items, term, resolvedOptions.filterOptions)
+        .sort sortItems(resolvedOptions.valueLookup)
+
+    @_ =
+      options: resolvedOptions
+      resultsCache: new hx.Map
+
+
+  clearCache: -> @_.resultsCache = new hx.Map
```
jmsmyth: Return type should be `this` or `undefined`

modules/autocomplete-feed/main/index.coffee
```diff
@@ -0,0 +1,99 @@
...
+          # Call the function then apply filtering
+          _.items(term, filterAndCallback)
+        else if term.length
+          # Apply filtering to the static object
+          filterAndCallback(_.items)
+        else
+          # Skip filtering and return the entire itemsset
+          cacheitemsAndCallback(_.items)
+
+  validateItems: (items) -> (hx.isArray(items) and items.length > 0) or hx.isFunction(items)
```
jmsmyth: What does `validateItems` do?

Would there be any issues with accepting an array of length 0? I'm thinking the picker could show the 'no results' message in this case. might have not fully understood what this is doing yet though, so ignore if that doesn't make sense..

modules/autocomplete-picker/main/index.coffee
```diff
@@ -0,0 +1,210 @@
...
+      this
+    else
+      @_.feed.items()
+
+  value: (value) ->
+    _ = @_
+    if arguments.length
+      _.valueText.text(_.options.loadingText)
+      _.feed.filter _.valueLookup(value), (results) =>
+        setPickerValue(this, results, 'api')
```
jmsmyth: Potential race condition here - I guess _.feed.filter is async? What happens when value is called multiple times in quick succession?

modules/autocomplete-picker/main/index.coffee
```diff
@@ -0,0 +1,210 @@
...
+      if resolvedOptions.value
+        @value(resolvedOptions.value)
+      else
+        valueText.text(resolvedOptions.chooseValueText)
+
+      if resolvedOptions.disabled
+        @disabled(resolvedOptions.disabled)
+
+
+  clearCache: -> @_.feed.clearCache()
```
jmsmyth: return types: `this` or `undefined`

modules/autocomplete-feed/main/index.coffee
c-frater: validateItems exists so the feed validates the items, rather than whatever uses the feed, mainly because the feed is the only part that should care about the data and it's validity.

The reasoning behind the length thing is because there's no point in having a picker if the length of the items is 0. I can't think of any reason you'd pass a fixed array in of length 0 for a legitimate use?

modules/autocomplete-feed/module.json
c-frater: I'll ditch them then 👍

modules/autocomplete-picker/main/index.coffee
```diff
@@ -0,0 +1,210 @@
...
+})
+
+validateItems = (feed, items) ->
+  if not feed.validateItems(items)
+    hx.consoleWarning "hx.AutocompletePicker: the items was expected to be an array of items or a function, you supplied: #{items}"
+    false
+  else
+    true
+
+setPickerValue = (picker, results, cause) ->
```
jmsmyth: glad to see the `cause` is sent with the events :+1:

modules/autocomplete-picker/main/index.coffee
```diff
@@ -0,0 +1,210 @@
...
+          if otherResults.length > 0
+            otherResults = [otherResultsItem].concat(otherResults)
+          renderMenu(results.concat(otherResults))
+
+      setValue = (item) =>
+        setPickerValue(this, [item], 'user')
+        menu.hide()
+
+      input = hx.detached('input').class('hx-autocomplete-picker-input')
+        .on 'input', (e) -> populateMenu(e.target.value)
```
jmsmyth: `populateMenu` calls `feed.filter` which is async, so the ongoing `feed.filter` queries should be cancelled (if there are any) whenever a new populateMenu call starts. Might mean the feed api needs tweaking to allow canceling?

modules/autocomplete-picker/main/index.coffee
c-frater: That is a very fair point. `feed.filter` is indeed async - the feed needs something to deal with multiple calls and the input should also be debounced

modules/autocomplete-picker/main/index.coffee
c-frater: It does indeed need tweaking :)

modules/autocomplete-feed/main/index.coffee
jmsmyth: The situation I was thinking of was: if you had something else that changes the values available to choose from in the picker, and you only have a small number of items, it could make sense in that case to set the items directly, rather than setting up a function feed. And in that case you _might_ sometimes want to set the items to an empty array. Its a bit contrived I admit, but the use case could happen

modules/autocomplete-picker/main/index.coffee
```diff
@@ -0,0 +1,210 @@
...
+        .on 'keydown', (e) ->
+          if input.value().length
+            if (e.which or e.keyCode) is 13 and menu.cursorPos is -1
```
jmsmyth: could define `13` as a named constant so it's easier to know which key this refers to

modules/autocomplete-picker/main/index.coffee
jmsmyth: And value could have a callback (or maybe return a promise - unless we want to leave that for #26)

c-frater: Value doesn't need a callback as you can listen to `on 'change'` where the cause is the API ;)

modules/autocomplete-feed/main/index.coffee
```diff
@@ -0,0 +1,104 @@
...
+
+
+  clearCache: ->
+    @_.resultsCache = new hx.Map
+    this
+
+  filter: (term = '', callback) ->
+    _ = @_
+    thisFilter = term + hx.randomId()
+    _.lastFilter = thisFilter
```
c-frater: Not sure if this is the best way of doing this?

modules/autocomplete-feed/test/spec.coffee
```diff
@@ -0,0 +1,258 @@
...
+      af = new hx.AutocompleteFeed
+      af.items(items).should.equal(af)
+      cb = chai.spy()
+      af.filter('a', cb)
+      af.filter('ab', cb)
+      af.filter('abc', cb)
+      clock.tick(4000)
+      cb.should.have.been.called.once()
+      cb.should.have.been.called.with(['abc'])
+      clock.restore()
```
c-frater: Race condition test :)

modules/autocomplete-picker/main/index.coffee
jmsmyth: Which `'change'` event corresponds to which call to value() though? there is no way of knowing

modules/autocomplete-feed/main/index.coffee
c-frater: By 'this' I mean the 'only callback with the last called filter'.
Everything is run and cached for every filter but the callback is only called when `thisFilter === _.lastFilter`

modules/autocomplete-picker/main/index.coffee
c-frater: Fair point, it's quite trivial to add so I might as well add in a callback.
I would use a Promise but as nothing else in Hexagon uses them as yet it probably does make sense to wait and do that properly with #26

modules/autocomplete-picker/main/index.coffee
```diff
@@ -0,0 +1,221 @@
...
+    else
+      @_.feed.items()
+
+  value: (value, callback) ->
+    _ = @_
+    if arguments.length
+      _.valueText.text(_.options.loadingText)
+      _.feed.filter _.valueLookup(value), (results) =>
+        setPickerValue(this, results, 'api')
+        callback?(value, results.length is 1)
```
c-frater: I realised that this isn't matching up with what setPickerValue is doing - It should return the first result if the results length is > 0, not the value passed in.

modules/autocomplete-feed/test/spec.coffee
jmsmyth: Could it be changed to make 'ab' take longer to respond than 'abc'?

modules/autocomplete-feed/main/index.coffee
```diff
@@ -0,0 +1,104 @@
...
+        _.resultsCache.set(term, {
+          results: results,
+          otherResults: otherResults
+        })
+        if thisFilter is _.lastFilter
+          callback(results, otherResults)
+
+    if _.options.matchType is 'external' and hx.isFunction(_.items)
+      # The matching is external so we don't filter here
+      _.items(term, cacheitemsAndCallback)
```
jmsmyth: does it use the cache in this case?

c-frater: Yes - it calls into `cacheItemsAndCallback` which is what AutoComplete did before (in a slightly more convoluted way)
The theory being if you're using external matching the search will return the same results for the same term. If that's not the case then calling `clearCache` in the external match function will provide the 'never cache' option. 
It might be worth adding an option called `noCache` or something along those lines?

jmsmyth: A `noCache` option sounds good.

Are you sure it is using the cache? - I might be reading it wrong, but `cacheitemsandCallback` only writes to the cache, never reads..

modules/autocomplete-feed/main/index.coffee
```diff
@@ -0,0 +1,107 @@
...
+  clearCache: ->
+    @_.resultsCache = new hx.Map
+    this
+
+  filter: (term = '', callback) ->
+    _ = @_
+    thisFilter = term + hx.randomId()
+    _.lastFilter = thisFilter
+
+    cacheitemsAndCallback = (results, otherResults = []) =>
```
jmsmyth: `cacheitemsAndCallback` -> `cacheItemsAndCallback`

modules/autocomplete-feed/main/index.coffee
c-frater: Oh yeah... I'm not entirely sure why but I thought I'd added the `if _.resultsCache.has...` block as the first step - It should be using the cache ;)
I'll also add the `noCache` option :)

modules/util/main/index.coffee
```diff
@@ -49,10 +49,11 @@ hx.supports = (name) ->
 hx.debounce = (duration, fn) ->
   timeout = undefined
   return ->
+    origArgs = arguments
     if timeout then clearTimeout(timeout)
     f = ->
       timeout = undefined
-      fn()
+      fn.apply(this, origArgs)
```
c-frater: Pass arguments through for `hx.debounce`

modules/menu/main/index.coffee
```diff
@@ -223,11 +223,16 @@ class Menu extends hx.EventEmitter
     @options.dropdownOptions.ddClass = 'hx-menu ' + if colorClass? then 'hx-' + colorClass else @options.dropdownOptions.ddClass
 
     dropdownContent = (node) ->
+      elem = hx.select(node)
+      menuItems = elem.select('.hx-menu-items')
+      if menuItems.empty()
+        menuItems = elem.append('div').class('hx-menu-items')
```
c-frater: Populate menu items inside the `hx-menu-items` div inside the dropdown. This is so we can add content outside the menu items and still re-render the menu items.

modules/autocomplete-feed/main/index.coffee
c-frater: +1 - I think `cacheItemsThenCallback` makes more sense - `and` makes it sound like it's caching the callback as well as the items...

c-frater:
This also resolves #209

jmsmyth:
@c-frater Could you provide an overview of the api & example of how this would be used?

c-frater:
The API works the same way as the picker, with the addition of extra options for the autocomplete functionality.

The api has the following methods:
- `clearCache` - clears the autocomplete cache
- `hide` - hides the dropdown
- `disabled` - gets/sets the disabled state of the component
- `items` - gets/sets the items (can be an async function)
- `value` - gets/sets the value
- `renderer` - gets/sets the renderer

The use case for this is when you have a long list of items to select from but still want to use a picker-style selector.

Generally I'd expect this to be used where the user is presented with a large set of options that they must select a value from (e.g. locales, timezones etc) but they know what option they are looking for (or at least roughly know)

This example illustrates the purpose. It consists of a list of 1000 'random' items that (mostly) have 5 characters:
```es6
const items = hx.range(1000).map((x) => (x / 1000).toString(36).replace(/[^a-z]+/g, '').substr(0, 5))

hx.select('body').clear()
  .add(hx.autocompletePicker(items))
```

In this example, if you wanted to select `xjveg`, you'd have to scroll through a lot of items to find it as it's near the end of the list. However, with the search box, it becomes much simpler.

The API also allows for results to be retrieved from a server asynchronously for the complete item set, as well as filtered results, in the same way as the existing autocomplete.

PR #200, opened 2016-04-08
Configurable user facing text

c-frater:
I've written a module for configuring the user facing text.

You can set up the text using:
```js
hx.userFacingText('noData', {'en': 'No Data'})
hx.userFacingText(['dataTable','noData'], {en: 'No Data'})
```

These can then be retrieved using the symmetrical getter:
```js
hx.userFacingText('noData')
hx.userFacingText(['dataTable','noData'])
```

You can also retrieve all the currently set strings using:
```js
hx.userFacingText()
```
Which is useful if you want to see what you need to localise.

The theory is that each module would set up it's own user facing text which would then be stored in the global `userFacingText` object then each module would internally use the function to retrieve the text.

The retrieval checks for text in the current `hx.preferences.locale()` and then falls back to `en` which _should_ always exist (as it will be set up in the module code).

### DIFF
--- a/modules/user-facing-text/main/index.coffee
+++ b/modules/user-facing-text/main/index.coffee
@@ -1,52 +1,42 @@
-class UserFacingText extends hx.EventEmitter
-  constructor: ->
-    super
-    @_ =
-      localisedText: {}
-
-  localisedText: (stringPath, locales) ->
-    if arguments.length
-      strings = if hx.isString(stringPath)
-        [stringPath]
-      else stringPath
-
-      if locales?
-        newText = tmp = {}
-        for string, i in strings
-          if i is strings.length - 1
-            # Merge the locales to prevent mutation
-            tmp[string] = hx.merge locales
-          else
-            tmp[string] = {}
-            tmp = tmp[string]
-
-        @_.localisedText = hx.merge.defined(@_.localisedText, newText)
-        @emit 'change'
-        this
-      else
-        localisedText = undefined
-        tmp = @_.localisedText
-        for string, i in strings
-          if i is strings.length - 1
-            localisedText = tmp[string]
-          else if tmp?
-            tmp = tmp[string]
-          else
-            break
-        if not localisedText
-          hx.consoleWarning 'hx.userFacingText.localisedText: No text was found for locale for the path ' + stringPath
-        else
-          localisedText[hx.preferences.locale()] || localisedText['en'] # There must be a fallback in english.
-    else
-      @_.localisedText
-
-
-globalUserFacingText = new UserFacingText
+_ =
+  localisedText: {}
 
 userFacingText = (stringPath, locales) ->
-  globalUserFacingText.localisedText.apply(globalUserFacingText, arguments)
+  if arguments.length
+    strings = if hx.isString(stringPath)
+      [stringPath]
+    else stringPath
+
+    if locales?
+      newText = tmp = {}
+      for string, i in strings
+        if i is strings.length - 1
+          # Merge the locales to prevent mutation
+          tmp[string] = hx.merge locales
+        else
+          tmp[string] = {}
+          tmp = tmp[string]
+      _.localisedText = hx.merge.defined(_.localisedText, newText)
+      this
+    else
+      textObject = undefined
+      tmp = _.localisedText
+      for string, i in strings
+        if i is strings.length - 1
+          textObject = tmp[string]
+        else if tmp?
+          tmp = tmp[string]
+        else
+          break
+      if not textObject
+        hx.consoleWarning 'hx.userFacingText.localisedText: No text was found for locale for the path ' + stringPath
+      else
+        textObject[hx.preferences.locale()] || textObject['en'] # There must be a fallback in english.
+  else
+    _.localisedText
 
 hx.userFacingText = userFacingText
+hx.userFacingText.emitter = userFacingTextEmitter
 
 # For tests
-hx.userFacingText._ = globalUserFacingText._
+hx.userFacingText._ = _
--- a/modules/user-facing-text/main/index.coffee
+++ b/modules/user-facing-text/main/index.coffee
@@ -13,9 +13,11 @@ completeGetterSetter = (object) ->
   else
     hx.clone _.localisedText
 
+isValid = (value) -> hx.isString(value) and value.length
+
 partialGetterSetter = (module, key, value) ->
-  if hx.isString(module) and module.length and hx.isString(key) and key.length
-    if hx.isString(value) and value.length
+  if isValid(module) and isValid(string)
+    if isValid(value)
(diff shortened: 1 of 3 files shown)


modules/user-facing-text/main/index.coffee
```diff
@@ -0,0 +1,45 @@
+_ =
+  localisedText: {}
```
c-frater: This is stored in _ so that the tests can use it safely

modules/user-facing-text/main/index.coffee
```diff
@@ -0,0 +1,45 @@
...
+    if locales?
+      newText = tmp = {}
+      for string, i in strings
+        if i is strings.length - 1
+          # Merge the locales to prevent mutation
+          tmp[string] = hx.merge locales
+        else
+          tmp[string] = {}
+          tmp = tmp[string]
+      userFacingTextEmitter.emit('change')
```
c-frater: I'm not sure why you'd need to detect this but I thought it might be a 'nice to have', easy enough to remove if it doesn't make sense.

modules/user-facing-text/main/index.coffee
```diff
@@ -0,0 +1,45 @@
...
+        if i is strings.length - 1
+          textObject = tmp[string]
+        else if tmp and tmp[string]
+          tmp = tmp[string]
+        else break
+
+      if not textObject
+        hx.consoleWarning 'hx.userFacingText.localisedText: No text was found for locale for the path ' + stringPath
+      else
+        textObject[hx.preferences.locale()] || textObject['en'] # There must be a fallback in english.
```
c-frater: Is it worth checking if the `hx.preferences.locale()` version exists then checking for the `en` version and then showing a warning if it doesn't exist or should there _always_ be an english version?
I guess the real question is 'Should people be configuring their own non-hexagon text with this module or is it just for configuring hexagon?'

c-frater:
The idea behind this is to get the module working for getting/setting the text and then implement it on the modules with user facing text. The first step is working out how it _should_ be done which is what this pull request covers. This module currently has no effect on any of the other modules, the modules should set up their own user facing text.

georgesimms:
* presumably the namespaced getter should be `hx.userFacingText(['dataTable','noData'])`, right ?
* Is there any excuse not to use the namespaced getter ?
* Given the trend to move away from positional arguments in arrays, maybe we should
```
hx.userFacingText({
  module: 'dataTable',
  key: 'noData'
})
```
* What does hx.userFacingText return when called with no arguments ? A plain object ? Keyed on what ?
* Is there scope for locale inheritence (eg en-GB inherits from en-US)

jmsmyth:
Some thoughts:

## Define per module rather than global
I'm wondering if it could make more sense to define this per module rather than having to key on multiple things `{moduleName, locale, key}`

For example

```
// to set
hx.DataTable.userFacingText({
  noData: 'No Data'
})

// to get
hx.DataTable.userFacingText()
```

The argument for this approach is that it would keep modules self contained, which is good for testing and code readability/reasoning. 

## Drop locale from the api
Not sure if the locale needs to be part of the api - this could be left for something else to handle _if_ multiple locales are needed. 

Something like this could work
```
  hx.preferences.on('localechange', () => {
    const locale = hx.preferences.locale()
    // set up all the localised user strings here...
    hx.DataTable.userFacingText({

    })
  })
```

## Drop locale from the api

Another thing, it would be nice if the user facing text could be set per instance as well as globally, to allow custom messages depending on the context:

```
const dataTable = new hx.DataTable(...)

// to set
dataTable.userFacingText({
  noData: 'No Users'
})

// to get
dataTable.userFacingText()
```

c-frater:
Define per module

I agree that it makes sense but that also means we'd have to implement it separately for every module, or create another module that isn't accessible to users that can be shared. 
I figured the name spacing allows us to configure per-module and also gives some flexibility in terms of creating custom 'tags' if required.


Drop the locale

I agree with the possibility of dropping the locale from the API - the user would have to create custom text anyway for any non-english locales.
The only thing I'd say is that in this case there would need to be a 'reset' method of some description or some way to get back to the defaults.

Per instance

The way I thought it would work was have the text remain in the options but the defaults be changed to use this module. That way every instance will be customisable if required. 


I think this is definitely one to discuss a bit more with regard to how it should be used - something that the user can add 'custom' tags to for setting text for their own fields or something that's purely for configuring the text within Hexagon. It's currently set up in a more customisable way.


I think I'm happy to change the arguments to something like `(module, key)` as that makes it easier to manage both internally and externally and dropping the locale would definitely make the setter/getter functionality simpler - my only concern is that we're then relying on people writing their own function(s) for switching locales which would be something standard that we could support

c-frater:
##Suggested API

Below is a suggestion taking into account the previous comments:

```js
// Singular Setter
hx.userFacingText(module: String, key: String, value: String) // No return value?

// Singular Getter
hx.userFacingText(module: String, key: String): String

// Complete Setter
hx.userFacingText(strings: Object) // No return value?

// Complete Getter
hx.userFacingText(): Object

// Get Defaults - Must be 'immutable' return value
hx.userFacingText.defaults(): Object
```

The complete object could look like this to make it easy to manage:
```js
{
  dataTable: {
    noData: 'No Data',
    ...// other keys
  }
  ...// other modules
}
```

Changes to hx.DataTable
```js
hx.userFacingText('dataTable', 'noData', 'No Data'); // Add at top of file / somewhere standard across modules - defined if module is included, not if it is called (safer / easier to manage / easier to override)

...

resolvedOptions = ...
  ...
  noDataMessage: function () { return hx.userFacingText('dataTable','noData') } // May not need to be a function...
  ...
```

- Removed array based keys
- Removed internal locale (could be added _relatively_ easily if required by changing the value to be an object and having an additional 'locale' argument?)
- Retain per-instance by keeping the 'option' for the text and just setting the default to use `userFacingText`
- Keeps the code for storing/handling the text in one module and just set/get in the other modules (easier to maintain)
- 'Complete' getter/setter works better and is easier to implement custom locales (i.e. calling the getter shows all the set text only for the modules you have included so you know what you need to localise)

modules/user-facing-text/main/index.coffee
```diff
@@ -0,0 +1,43 @@
...
+    else
+      hx.consoleWarning "hx.userFacingText: The value provided must be a string but was passed value: #{value}"
+  else
+    hx.consoleWarning "hx.userFacingText: A module and key are expected as strings but was passed module: #{module} and key: #{key}"
+
+userFacingText = ->
+  if hx.isObject(arguments[0]) or arguments.length is 0
+    completeGetterSetter.apply(this, arguments)
+  else
+    partialGetterSetter.apply(this, arguments)
```
c-frater: Split into two functions to make it easier to follow the logic - there was a lot of nesting to handle the 'complete' getter setter

PR #208, opened 2016-04-18
Add the Tooltip component

### LINKED ISSUE #16: New Component: Tooltip

jmsmyth:
resolves #16


jmsmyth:
So, this is a _pretty_ basic implementation - mostly copied over from the incubator. Is it sufficient for an initial version?

modules/tooltip/main/index.coffee
```diff
@@ -0,0 +1,21 @@
+
+hx.tooltip = (selector, opts) ->
+
+  options = hx.merge({
+    message: ''
```
c-frater: Should message be an argument as opposed to an option?

modules/tooltip/main/index.coffee
```diff
@@ -0,0 +1,21 @@
...
+    hx.select(element)
+      .text(options.message)
+      .classed('hx-tooltip', true)
+
+  dropdownOptions = {
+    mode: 'hover',
+    align: options.align
+  }
+
+  new hx.Dropdown(hx.select(selector).node(), content, dropdownOptions)
```
c-frater: Would it not make more sense to rely on the dropdown for the string renderer?
```
hx.Dropdown(selector, options.message, {ddClass: 'hx-tooltip'})
```

c-frater: This could potentially be an extension to dropdown as an option that adds the 'hx-tooltip' class for you when passing in a string as the `dropdownContent` - It doesn't feel like it's _different_ enough from a dropdown to be called a tooltip...

jmsmyth: The string option for dropdown is interpreted as html, which something we should really deprecate and avoid in the future because of the security implications. Allowing the user to pass in a text-only string or a renderer function would make sense - but that would still need a custom content renderer function here until dropdown is updated - which can't happen until 2.0.0

I see what you mean about there not being much difference between dropdown and this. Could this be because dropdown just happens to have everything that tooltip needs to work _at the moment_. Extending dropdown would imply that this will always work using a dropdown, which might not always be the case (although probably will be :-p). 

I think it's better to keep things as separate modules as much as possible - even if it does use another module under the hood, the user doesn't need to know that - this gives us more flexibility to change how it works without affecting the api and the preconceptions the user has

c-frater:
Having had a look at the api, I'm not sure the current implementation is much more useful than just using a regular dropdown. The main issue is that setting the `align` option to anything other than 'right' looks odd (as there is currently left margin), this could be resolved by #250 and relying on that. 

It might make more sense to expose the positioning logic for dropdowns and create a new component that shares that part of the dropdown code, rather than using the dropdown itself.

Issue #201, opened 2016-04-11
Tag Input: add an option for 'placeholder'


c-frater:
Also make sure it propagates through form builder - It probably makes sense to be able to do 
```js
var form = new hx.Form(hx.select(elem).append('form').node())
     .addTagInput('thing', {placeholder: 'Placeholder'})
```

PR #220, opened 2016-04-22
User facing text

### LINKED ISSUE #211: User Facing Text: convert all modules to use new user facing text module
TODO:
- [x] autocomplete
- [x] data-table
- [x] file-input
- [x] form
- [x] picker
- [x] plot
- [x] preferences
- [x] tag-input

c-frater:
Resolves #211
Add user facing text to modules

### DIFF
--- a/modules/file-input/main/index.coffee
+++ b/modules/file-input/main/index.coffee
@@ -1,9 +1,9 @@
 hx.userFacingText({
   fileInput: {
-    noFile: 'No File Chosen',
-    filesSelected: 'Files Selected',
-    chooseFile: 'Choose File'
+    chooseFile: 'Choose File',
     chooseFiles: 'Choose Files'
+    filesSelected: 'Files Selected: $numFiles',
+    noFile: 'No File Chosen',
   }
 })
 
@@ -49,9 +49,10 @@ class FileInput extends hx.EventEmitter
       multiple: false
       dragEnabled: true
       buttonClass: 'hx-action'
-      noFilesText: hx.userFacingText('fileInput', 'noFiles')
-      filesSelectedText: hx.userFacingText('fileInput', 'filesSelected')
+
       buttonText: hx.userFacingText('fileInput', 'chooseFile')
+      filesSelectedText: hx.userFacingText('fileInput', 'filesSelected')
+      noFilesText: hx.userFacingText('fileInput', 'noFile')
 
     resolvedOptions = hx.merge defaults, options
 
@@ -133,9 +134,11 @@ class FileInput extends hx.EventEmitter
           dropdown.hide() if dropdown.isOpen()
           selectedFiles.append filePreview(fileMap.values()[0])
         else
+          localizedLength = length.toLocaleString(hx.preferences.locale())
+          filesSelectedText = resolvedOptions.filesSelectedText.replace('$numFiles', localizedLength)
           selectedFiles
             .classed 'hx-btn', true
-            .add hx.section().text "#{length} #{resolvedOptions.filesSelectedText}"
+            .add hx.section().text filesSelectedText
             .add hx.detached('i').class 'hx-file-input-dropdown-icon hx-icon hx-icon-chevron-down'
             .on 'click', 'hx.file-input', -> dropdown.show()
       else
--- a/modules/file-input/main/index.coffee
+++ b/modules/file-input/main/index.coffee
@@ -1,3 +1,13 @@
+hx.userFacingText({
+  fileInput: {
+    noFile: 'No File Chosen',
+    filesSelected: 'Files Selected',
+    chooseFile: 'Choose File'
+    chooseFiles: 'Choose Files'
+  }
+})
+
+
 getFileUID = (file) -> file.name + file.size + file.lastModified + file.type
 
 fileValidator = (file, acceptedExtensions) ->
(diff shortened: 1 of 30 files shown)


modules/file-input/test/spec.coffee
```diff
@@ -16,116 +18,265 @@ describe "File Input", ->
...
+    # fake some dom event stuff
+    stopPropagation: ->
+  }
 
-  beforeEach ->
+  before ->
     hx.consoleWarning = chai.spy()
-    fixture.clear()
+    hx._.dropdown.attachToSelector = '#fixture'
+    clock = sinon.useFakeTimers()
```
c-frater: Came across a random issue where 2 of the data table event tests were failing.
Defining the spies / clock etc. in the before function seemed sensible anyway - probably worth doing this everywhere?

modules/form/test/spec.coffee
```diff
@@ -0,0 +1,5 @@
+describe 'form', ->
+  it 'should have user facing text defined', ->
+    hx.userFacingText('form','missingRadioValue').should.equal('Please select one of these options')
+    hx.userFacingText('form','missingValue').should.equal('Please fill in this field')
+    hx.userFacingText('form','typeMismatch').should.equal('Please enter a valid value for this field')
```
c-frater: Added tests for the new user facing text - makes sure the dependency is added in the right place and that the text exists / hasn't changed

c-frater:
This isn't 'done' yet - I just came across an interesting issue with data table (see comments) and figured I'd push what I've got anyway

modules/file-input/test/spec.coffee
```diff
@@ -16,116 +18,265 @@ describe "File Input", ->
...
+    hx.userFacingText('fileInput','chooseFiles').should.equal('Choose Files')
+    hx.userFacingText('fileInput','filesSelected').should.equal('Files Selected: $numFiles')
+    hx.userFacingText('fileInput','noFile').should.equal('No File Chosen')
+
 
-  emitEvent = (node, event, data) ->
-    hx.select.getHexagonElementDataObject(node).eventEmitter
-      .emit event, data
+  # Not sure how to test this?
+  # it 'should load a preview image', ->
```
c-frater: It uses a `FileReader` and I'm not sure how to create something that a file reader can read? I tried using a `Blob` but node doesn't have those...

modules/file-input/test/spec.coffee
```diff
@@ -16,116 +18,265 @@ describe "File Input", ->
...
+      fakeDragEvent.preventDefault.should.have.been.called.once()
+      fakeDragEvent.stopPropagation.should.have.been.called.once()
+
+      fakeNodeEvent(sel.node(), 'dragover')(fakeDragEvent)
+      fakeDragEvent.preventDefault.should.have.been.called.twice()
+      fakeDragEvent.stopPropagation.should.have.been.called.twice()
+
+      fakeNodeEvent(sel.node(), 'drop')(fakeDragEvent)
+      fakeDragEvent.preventDefault.should.have.been.called.exactly(3)
+      fileInput.value().should.eql([ fakeFile1 ])
```
c-frater: These tests are assuming it's reasonable to mimic the drag events using the 'fake node event' function

modules/form-builder/module.json
```diff
@@ -1,15 +1,12 @@
...
+    "picker",
     "selection",
-    "event-emitter",
     "tag-input",
-    "picker"
-  ],
-  "keywords": [
-    "form",
-    "builder"
+    "util"
```
c-frater: Remove 'keywords' and sort dependencies alphabetically - done this in every module.json I've looked in because I figure it's a reasonable thing to do :)

modules/form/module.json
```diff
@@ -1,6 +1,7 @@
 {
-  "dependencies": [],
-  "keywords": [
-    "form"
+  "dependencies": [
+    "selection",
+    "user-facing-text",
+    "util"
```
c-frater: Add dependencies to form - It didn't have any and needed them!

c-frater:
This isn't 'done' yet - I just came across an interesting issue with data table (see comments) and figured I'd push what I've got anyway

c-frater:
I've added user facing text to all the modules that require it.
Some modules have had additional options added where one didn't exist previously and these will need to be documented

modules/file-input/main/index.coffee
```diff
@@ -39,14 +49,15 @@ class FileInput extends hx.EventEmitter
       multiple: false
       dragEnabled: true
       buttonClass: 'hx-action'
-      noFilesText: 'No File Chosen'
-      filesSelectedText: 'Files Selected'
-      buttonText: 'Choose File'
+
+      buttonText: hx.userFacingText('fileInput', 'chooseFile')
```
jmsmyth: the difference in naming here might be confusing

jmsmyth:
Looks fine to me

modules/file-input/main/index.coffee
c-frater: Generally I've gone with what the text is describing as the variable name, rather than where it will be used, I figured it would make more sense than having a localisation object that just says 'buttonText' - it's not clear what the button _does_ so it's not clear what text should be put in there

jmsmyth: I was more getting at the use of `buttonText`/`chooseFile` for the same option (depending on if you set it via `hx.userFacingText` or the `options` object).

However backwards compatibility is obviously needed, so just something to revisit in 2.0.0..

c-frater:
This also resolves #201 - The option was added as part of the conversion to use user facing text

PR #231, opened 2016-04-27
Add node version 6 to .travis.yml

jmsmyth:


jmsmyth:
All the tests actually pass now - only the build doesn't terminate in travis for some reason..

c-frater:
Tried locally with node 6.2.0 
It looks like `karma` never calls the 'callback' when it's done here:

```js
return new Promise(function (resolve, reject) {
    console.log('startKarma')
    karma.server.start(cfg, function (exitStatus) {
      console.log('exitStatus', exitStatus)
      resolve(exitStatus)
    })
  })
```

It never logs 'exitStatus' - not sure if this is a config related issue (reporter?) or karma

c-frater:
The current status of this PR is that Karma is not exiting correctly. 
We're testing periodically with node 6+ until it is supported - it will stay in review until karma has updated.

c-frater:
It looks like this may have been an issue with node as opposed to Karma, running using node version 6.2.1 passes the tests. This can probably be merged now :)

PR #251, opened 2016-06-16
Preferences: prefer navigator.languages over navigator.language

georgesimms:
On Chrome, navigator.language is determined at the point chrome is downloaded and is unaffected by re-ordering the languages in the chrome settings.

`navigator.languages`, however, are.

See [here](https://alicoding.com/detect-browser-language-preference-in-firefox-and-chrome-using-javascript/) for more details.

### DIFF
--- a/modules/preferences/main/index.coffee
+++ b/modules/preferences/main/index.coffee
@@ -207,7 +207,7 @@ class Preferences extends hx.EventEmitter
       modal: modal
     }
 
-    defaultLocaleId = moment?.locale() or navigator.language
+    defaultLocaleId = moment?.locale() or navigator.languages?[0] or navigator.language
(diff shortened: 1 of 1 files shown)


c-frater:
As a note, `navigator.language` is generally affected initially by the OS, as opposed to the browser. This is more of an issue when installing in the UK as the majority of the time, the OS is set up with `en-US` as the default (which is particularly the case when using a Mac). 
Regardless of the setup, `navigator.languages` is the better choice as it is at browser level which means that someone may have another language set as their preferred language despite working on a machine with `en-US` as the language.

Also, the change looks good to me

georgesimms:
Mergeable ? If so let's see if we can release it ?

jmsmyth:
Yes, can we merge this please?

PR #253, opened 2016-06-21
Changed card title element type to span (from div)

### LINKED ISSUE #245: Card: change hx.card.title() to create a span rather than a div
`hx.card.title()` creates a `<div>` with the `hx-card-title hx-header` classes. Instead, it should probably create a `<span>` so that you can put icons next to your titles without the title wrapping onto the next line.

The workaround I am using in the meantime is:

```
hx.detached('span').class('hx-card-title hx-header').text('Title')
```

in place of

```
hx.card.title({text: 'Title'})
```

jmsmyth:
Resolves #245


c-frater:
Looks good to me 👍

Issue #222, opened 2016-04-25
Preferences: incorrectly named variable in timezoneOffsetLookup


c-frater:
The variable name used in the default timezoneOffsetLookup function is called `datestamp` in the function signature but used as `timestamp`
```js
@timezoneOffsetLookup (timezone, datestamp) ->          // datestamp
        -(moment.tz.zone(timezone).offset(timestamp) / 60) // timestamp
```

Workaround:
```js
hx.preferences.timezoneOffsetLookup(function (timezone, timestamp) {
  return -(moment.tz.zone(timezone).offset(timestamp) / 60);
})
```

PR #255, opened 2016-06-27
fixed issue #222

jonthans:
make hexagon locale the canonical source for locale and make moment
shadow hexagon.

### DIFF
--- a/modules/preferences/main/index.coffee
+++ b/modules/preferences/main/index.coffee
@@ -219,7 +219,7 @@ class Preferences extends hx.EventEmitter
 
     defaultLocaleId = navigator.languages?[0] or navigator.language
 
-    @on('localechange', (value) -> moment.locale(value))
+    @on('localechange', (value) -> moment?.locale(value))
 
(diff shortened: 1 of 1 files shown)


jmsmyth:
LGTM 👍

modules/preferences/main/index.coffee
```diff
@@ -217,15 +217,18 @@ class Preferences extends hx.EventEmitter
       modal: modal
     }
 
-    defaultLocaleId = moment?.locale() or navigator.languages?[0] or navigator.language
+    defaultLocaleId = navigator.languages?[0] or navigator.language
+
+    @on('localechange', (value) -> moment?.locale(value))
```
c-frater: Rather than adding a listener to itself, this should instead be done when the locale is set in the `locale` method (i.e. when the `localechange` event is emitted)

c-frater:
This fixed #222

modules/preferences/main/index.coffee
georgesimms: Actually, should we only attach this listener if moment is there ?

PR #257, opened 2016-06-28
fix #232

### LINKED ISSUE #232: Fluid: The fluid hx.notice.head() function uses the wrong class.
The `hx.notice.head()` function creates an element with the `hx-notice-head` class, however the notice module uses `hx-notice-header`. Either the fluid api should be changes to use `hx-notice-header` or notice should use `hx-notice-head`.

In either case, adding `hx-notice-head` as a class for the notice module whilst retaining the existing class will make this a pre-2.0.0 change.

c-frater:
deprecate 'hx-notice-header' in favour of 'hx-notice-head'

### DIFF
--- a/modules/notice/main/index.scss
+++ b/modules/notice/main/index.scss
@@ -1,6 +1,7 @@
 .hx-notice {
   margin: 0.4em 0;
 
+  // XXX 2.0.0 - hx-notice-header should be removed
   .hx-notice-head,
   .hx-notice-header {
     padding: 0.4em;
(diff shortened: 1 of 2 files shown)


modules/notice/main/index.scss
```diff
@@ -1,6 +1,7 @@
 .hx-notice {
   margin: 0.4em 0;
 
+  .hx-notice-head,
   .hx-notice-header {
```
jmsmyth: Might want to add a XXX [2.0.0] to remind us to remove the header version in 2.0.0

PR #259, opened 2016-06-30
Fix misnamed autocomplete option

georgesimms:
This bug causes asynchronous autocompletes not to work - this change should fix that.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -192,8 +192,14 @@
         <div class="example">
           <h3>Autocomplete</h3>
           <input id="autocomplete"></input>
+          <input id="autocomplete-async"></input>
           <script>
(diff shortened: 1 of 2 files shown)


demo/index.html
```diff
@@ -192,8 +192,14 @@
...
           <h3>Autocomplete</h3>
           <input id="autocomplete"></input>
+          <input id="autocomplete-async"></input>
           <script>
             new hx.AutoComplete('#autocomplete', ['Dave', 'Jane', 'Bob', 'Dan', 'Steve', 'Alejandro'])
+            new hx.AutoComplete('#autocomplete-async', function (term, callback) {
+              setTimeout(function () {
+                callback(['Bob', 'Kate', 'Steve', 'Ganesh', 'Lazlo'])
+              }, 1000)
+            })
```
georgesimms: This is mostly just to double check that this is indeed the cause of the problem, and that this change does fix it.

PR #260, opened 2016-06-30
Bugfix

c-frater:
Following a bugfix in autocomplete, I checked over a few of the other files and found some issues.

I also found multiple issues in older versions of iOS Safari to do with flex styles that caused issues when rendering pages.

- Added support for older versions of iOS
- Fixed another misnamed variable in autocomplete
- Corrected the spelling of a variable in preferences

### DIFF
--- a/modules/autocomplete/main/index.coffee
+++ b/modules/autocomplete/main/index.coffee
@@ -131,7 +131,7 @@ buildAutoComplete = (searchTerm, fromCallback, loading) ->
     if not filteredData?
       message.text = @options.loadingMessage
     else if searchTerm.length < @options.minLength
-      message.text = @options.minLengthMessage.replace('$minLength', @options.minLength)
+      message.text = @options.pleaseEnterMinCharactersMessage.replace('$minLength', @options.minLength)
     else if (searchTerm.length > 0 or @options.showAll) and filteredData.length is 0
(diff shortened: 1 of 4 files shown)


modules/autocomplete/main/index.coffee
```diff
@@ -131,7 +131,7 @@ buildAutoComplete = (searchTerm, fromCallback, loading) ->
     if not filteredData?
       message.text = @options.loadingMessage
     else if searchTerm.length < @options.minLength
-      message.text = @options.minLengthMessage.replace('$minLength', @options.minLength)
+      message.text = @options.pleaseEnterMinCharactersMessage.replace('$minLength', @options.minLength)
```
jmsmyth: Why has this changed?

jmsmyth: Isn't it the docs that are wrong in this case?

c-frater: Possibly - although I think the option was added in 1.4.0 so it would be a bugfix

c-frater: Checked, it was added as part of the update to userFacingText:
https://github.com/ocadotechnology/hexagonjs/pull/220/files#diff-d62945ddda1e2d4c4438d9a9defcb6c8R134

The option added in the function did not match the value in the constructor... Probably best to fix it this way than to swap the constructor argument

PR #262, opened 2016-07-07
Fix legends in pie charts

georgesimms:


modules/plot/main/coffee/Expose.coffee
```diff
@@ -24,7 +24,6 @@ hx.StraightLineSeries = StraightLineSeries
 hx.pieChart = (options) ->
   selection = hx.detached('div')
   pieChart = new PieChart(selection.node(), options)
-  pieChart.render()
```
georgesimms: There is no point rendering here, as the div has 0 height and width

modules/plot/main/coffee/Utils.coffee
```diff
@@ -293,7 +293,7 @@ populateLegendSeries = (selection, series) ->
         .attr('x', 0)
         .attr('width', 10)
         .attr('height', 10)
-        .attr('fill', if s.legendColor then s.legendColor() else s.color) # XXX: the else s.color is there for PieCharts... which should be fixed when pie charts are refactored
+        .attr('fill', if s.legendColor then s.legendColor() else s.fillColor) # XXX: the else s.color is there for PieCharts... which should be fixed when pie charts are refactored
```
georgesimms: Fix for pie chart legends - without this they are always black rather than the actual comment.

I am not sure entirely what the comment is referring to re refactoring

modules/plot/main/coffee/charts/PieChart.coffee
```diff
@@ -18,6 +18,8 @@ class PieChart extends hx.EventEmitter
 
   constructor: (@selector, options) ->
     super
+    hx.select @selector
+      .on 'resize', 'hx.plot', => @render()
```
georgesimms: This is where we set up the resize listener, so that when a detached div is added to the page the pie chart renders correctly.

modules/plot/main/coffee/charts/PieChart.coffee
```diff
@@ -79,7 +81,7 @@ class PieChart extends hx.EventEmitter
 
   render: ->
     self = this
-    selection = hx.select(@selector)
+    selection = hx.select(@selector).class('hx-pie-chart')
```
georgesimms: Give it 100% height and width - this is for pie charts that appear in detached elements that get added to the dom

Otherwise width and height are 0 below

c-frater: Thought about this and I'm not sure this is something we want to add really as it breaks the idea behind how graphs and charts work currently. Generally, you have to either set explicit heights for each individual graph/chart or you have a class that sets min/max values.

I think it would be better in this case to add:
```
hx.pieChart(options).style('height', '500px').style('width','200px')
hx.pieChart(options).class('myPieClass')
```

Combining this with the resize events in this PR would resolve the issues with the pie chart fluid api

c-frater: @georgesimms This PR is waiting for the styles for `.hx-pie-chart` to be removed (the class can still be applied to the node)

jmsmyth: The class should be applied with `classed('hx-pie-chart', true)` to avoid removing existing classes

demo/index.html
```diff
@@ -1402,7 +1403,23 @@
...
+              segments: hx.range(6).map(function (i) {
+                return {
+                  name: "Segment " + i,
+                  size: i + 1,
+                  fillColor: hx.cycle(hx.theme.plot.colors, i)
+                }
+              })
+            }
+            hx.select('#pie').style('height', '100px').add(
+              hx.pieChart({legendEnabled: true, data: data}).style('width', '100%').style('height', '100%')
```
c-frater: This probably doesn't need changing for the PR but you _shouldn't_ need `width` here as `div`s are 100% width by default (`display: block`)
I'd also suggest that the `height: 100px` be applied directly to the newly created `hx.pieChart` as in the majority of cases that's how it would be used

PR #263, opened 2016-07-07
Add hx.isNumber

### LINKED ISSUE #240: Util: add hx.isNumber
For the sake of completeness, parallel with hx.isString etc.

Something like

```
hx.isNumber  = (x) => typeof x === 'number' || x instanceof Number
hx.isNumber(NaN) // => true
```

jmsmyth:
Resolves #240 (a request to add hx.isNumber)

### DIFF
--- a/modules/util/main/index.coffee
+++ b/modules/util/main/index.coffee
@@ -141,7 +141,7 @@ hx.isObject = (obj) -> typeof obj is 'object' and not hx.isArray(obj) and obj is
 
 hx.isBoolean = (x) -> x is true or x is false or typeof x is 'boolean'
 
-hx.isNumber = (x) -> typeof x is 'number' || x instanceof Number
+hx.isNumber = (x) -> typeof x is 'number' or x instanceof Number
 
 # Not plain objects:
 # - Anything created with new (or equivalent)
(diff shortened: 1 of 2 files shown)


modules/util/main/index.coffee
```diff
@@ -141,6 +141,8 @@ hx.isObject = (obj) -> typeof obj is 'object' and not hx.isArray(obj) and obj is
 
 hx.isBoolean = (x) -> x is true or x is false or typeof x is 'boolean'
 
+hx.isNumber = (x) -> typeof x is 'number' || x instanceof Number
```
c-frater: as it's coffee, should this not be 'or' instead of '||'? :)

modules/util/test/spec.coffee
```diff
@@ -303,6 +303,23 @@ describe "Util", ->
     hx.isPlainObject(null).should.equal(false)
     hx.isPlainObject(undefined).should.equal(false)
 
+  it 'isNumber: should work', ->
+    class A
+      constructor: ->
+
+    hx.isNumber(new A).should.equal(false)
```
c-frater: Could also do `hx.isNumber(A)`? Although I guess this is probably equivalent to `hx.isNumber(->)`

jmsmyth: Yep, would be the same as `->`

c-frater:
LGTM

c-frater:

PR #265, opened 2016-07-13
Build hexagon with es modules

### LINKED ISSUE #36: Make hexagon modules use a proper module system.
So that they can be `imported`, or `required` as individual modules, which will make more sense as http/2 becomes more available.

Some investigation should be done to figure out which module system to use, that is compatible with coffee and es6 (probably commonJS?) to allow exporting each module with a variety of module formats. It would be ideal if file types could be mixed so that modules could be converted over to es6 over time.

Hexagon bundles could still be built, but using something like webpack instead of the existing solution (which does work, but using something standard is probably a better thing for the project)
- [x] set
- [x] map
- [x] list
- [x] util
- [x] event-emitter
- [x] color
- [x] selection
- [x] transition
- [x] interpolate
- [x] animate
- [x] pointer-events
- [x] icon
- [x] spinner
- [x] view
- [x] sort
- [x] component
- [x] morphs
- [x] click-detector
- [x] base
- [x] modal
- [x] notify
- [x] filter
- [x] form
- [x] dropdown
- [x] collapsible
- [x] palette
- [x] format
- [x] button
- [x] autocomplete
- [x] resize-events
- [x] menu
- [x] input-group
- [x] date-localizer
- [x] preferences
- [x] number-picker
- [x] layout
- [x] drag-container
- [x] autocomplete-feed
- [x] progress-bar
- [x] sticky-table-headers
- [x] picker
- [x] button-group
- [x] plot
- [x] ~~request~~
- [x] table
- [x] date-picker
- [x] time-picker
- [x] tag-input
- [ ] ~~morph-section~~
- [x] titlebar
- [x] slider
- [x] card
- [x] color-picker
- [x] color-scale
- [x] crumbtrail
- [ ] data-table
- [x] date-time-picker
- [x] ~~drawing~~
- [x] error-pages
- [x] fast-click
- [x] fluid
- [ ] form-builder
- [ ] inline-editable
- [ ] inline-picker
- [x] label
- [x] logo
- [x] ~~meter~~
- [x] notice
- [x] paginator
- [ ] pivot-table
- [ ] side-collapsible
- [ ] sidebar
- [ ] tabs
- [x] time-slider
- [x] toggle
- [ ] tree

jmsmyth:
Still a work in progress, so don't merge yet.

What is working so far:
- mixed coffeescript and es6 transpilation
- commonjs modules
- running the tests from a webpage or from the command line

In theory, this can be a minor release at the moment, but we should review again when the modules have all been converted over

Resolves #36


jmsmyth:
The progess can also be tracked in this issue #36

modules/animate/main/index.coffee
```diff
@@ -1,8 +1,13 @@
-### istanbul ignore next: ignore coffeescript generated code that can't be covered  ###
+transition = require('modules/transition/main')
+EventEmitter = require('modules/event-emitter/main')
+interpolate = require('modules/interpolate/main')
+select = require('modules/selection/main')
+hxMap = require('modules/map/main')
```
c-frater: Does this need to be `hxMap` because there's a global `map` or is it so it's obvious that it's not `Array::map`?

jmsmyth: There shouldn't be a global map, so it could be renamed to just map. I kind of wanted it to be `Map` rather than `map` so its clear it's a type that needs creating with new, but that would also be confusing since it isn't really a Map. Maybe HMap would be a bit clearer?

Issue #252, opened 2016-06-20
Issue when setting compact and sortEnabled together


ZRonchy:
When creating the data table, I mark compact as true and sortEnabled as as false, but the following class still displays on the page, causing a blank line showing on top of the table. class="hx-data-table-sort-control hx-data-table-sort-visible"

c-frater:
This should be resolved in #267

PR #267, opened 2016-07-20
Data table multi column filtering

### LINKED ISSUE #49: Data Table: extend the filter/search capability to include per column filtering
The current filter does not allow per-column filtering. 
A solution that allows this whilst keeping the search box style would be ideal as it works nicely in terms of user experience. 

Something like a control panel that the user can open/close that allows filtering per-column would be good.

This will most likely also involve moving the search box above the table, moving the pagination may also make sense.

c-frater:
Closes #49 

#### Table with advanced search enabled
screen shot 2016-07-20 at 15 18 25

#### Table with only advanced search
screen shot 2016-07-20 at 15 19 57

#### Advanced Search filter grouping
screen shot 2016-07-20 at 15 19 35

#### Advanced Search 'off' state 
screen shot 2016-07-20 at 15 19 41

#### Table with search above content
screen shot 2016-07-20 at 15 19 20

#### Compact table with top/bottom control panel
screen shot 2016-07-20 at 15 18 58

#### Compact table with top/bottom control panel
screen shot 2016-07-20 at 15 18 58

#### Compact table with top control panel open
screen shot 2016-07-20 at 15 19 12

### DIFF
--- a/modules/data-table/main/index.coffee
+++ b/modules/data-table/main/index.coffee
@@ -1,6 +1,7 @@
 hx.userFacingText({
   dataTable: {
     addFilter: 'Add Filter',
+    addFilterGroup: 'Add Filter Group',
     advancedSearch: 'Advanced Search',
     and: 'and',
     anyColumn: 'Any column'
@@ -10,6 +11,7 @@ hx.userFacingText({
     noData: 'No Data',
     noSort: 'No Sort',
     or: 'or',
+    removeFilterGroup: 'Remove Filter Group',
     rowsPerPage: 'Rows Per Page',
     search: 'Search',
     selectedRows: '$selected of $total selected.',
@@ -26,6 +28,18 @@ columnOptionLookup = (options, name, id) ->
     options.columns[id][name]
   else
     options[name]
+
+splitArray = (array, index) ->
+  left = if index is 0
+    []
+  else
+    array[0..index - 1]
+  right = if index is array.length - 1
+    []
+  else
+    array[index+1..array.length]
+  [left, array[index], right]
+
 
 class DataTable extends hx.EventEmitter
   constructor: (selector, options) ->
@@ -75,6 +89,8 @@ class DataTable extends hx.EventEmitter
       addFilterText: hx.userFacingText('dataTable', 'addFilter')
       clearFiltersText: hx.userFacingText('dataTable', 'clearFilters')
       anyColumnText: hx.userFacingText('dataTable', 'anyColumn')
+      removeFilterGroupText: hx.userFacingText('dataTable', 'removeFilterGroup')
+      addFilterGroupText: hx.userFacingText('dataTable', 'addFilterGroup')
 
       advancedSearchPlaceholder: hx.userFacingText('dataTable', 'search')
     }, options)
@@ -141,7 +157,7 @@ class DataTable extends hx.EventEmitter
         typePickerSel.component()
           .on 'change', (data) ->
             if data.cause is 'user'
-              prevFilters = dataTable.filter()
+              prevFilters = dataTable.advancedSearch()
               [leftFilterGroups, filterGroup, rightFilterGroups] = splitArray(prevFilters, filterGroupIndex)
 
               newFilters = if data.value.value is 'or'
@@ -151,7 +167,7 @@ class DataTable extends hx.EventEmitter
                 [leftAllButLast..., leftLast] = leftFilterGroups
                 [leftAllButLast..., [leftLast..., filterGroup...], rightFilterGroups...]
 
-              dataTable.filter(newFilters)
+              dataTable.advancedSearch(newFilters)
 
         anyColumn = {
           text: resolvedOptions.anyColumnText
@@ -177,51 +193,39 @@ class DataTable extends hx.EventEmitter
         columnPickerSel.component()
           .on 'change', (data) ->
             if data.cause is 'user'
-              prevFilters = dataTable.filter()
+              prevFilters = dataTable.advancedSearch()
               [leftFilterGroups, filterGroup, rightFilterGroups] = splitArray(prevFilters, filterGroupIndex)
               [leftFilters, filter, rightFilters] = splitArray(filterGroup, trueIndex)
               newFilter = hx.merge(filter, {
                 column: data.value.value
               })
-              dataTable.filter([leftFilterGroups..., [leftFilters..., newFilter, rightFilters...], rightFilterGroups...])
+              dataTable.advancedSearch([leftFilterGroups..., [leftFilters..., newFilter, rightFilters...], rightFilterGroups...])
 
         debouncedInput = hx.debounce 200, (e) ->
-          prevFilters = dataTable.filter()
+          prevFilters = dataTable.advancedSearch()
           [leftFilterGroups, filterGroup, rightFilterGroups] = splitArray(prevFilters, filterGroupIndex)
           [leftFilters, filter, rightFilters] = splitArray(filterGroup, trueIndex)
           newFilter = hx.merge(filter, {
             term: e.target.value
           })
-          dataTable.filter([leftFilterGroups..., [leftFilters..., newFilter, rightFilters...], rightFilterGroups...])
+          dataTable.advancedSearch([leftFilterGroups..., [leftFilters..., newFilter, rightFilters...], rightFilterGroups...])
 
         termInput = hx.detached('input').attr('placeholder', resolvedOptions.advancedSearchPlaceholder)
           .class('hx-data-table-advanced-search-input hx-section')
           .on 'input', debouncedInput
 
-        splitArray = (array, index) ->
-          left = if index is 0
-            []
-          else
-            array[0..index - 1]
-          right = if index is array.length - 1
-            []
-          else
-            array[index+1..array.length]
-          [left, array[index], right]
-
-
         removeBtn = hx.button({context: 'negative'})
           .add(hx.icon({class: 'hx-icon hx-icon-close'}))
           .on 'click', ->
-            prevFilters = dataTable.filter()
+            prevFilters = dataTable.advancedSearch()
             [leftFilterGroups, filterGroup, rightFilterGroups] = splitArray(prevFilters, filterGroupIndex)
             [leftFilters, _, rightFilters] = splitArray(filterGroup, trueIndex)
 
-            newFilters = if trueIndex is 0
+            newFilters = if trueIndex is 0 and filterGroupIndex is 0
+              [rightFilters, rightFilterGroups...]
+            else if trueIndex is 0
               [leftFilterGroup..., leftFilterGroupLast] = leftFilterGroups
-
               [_, filters...] = filterGroup
-
               newFilterGroup = [leftFilterGroupLast..., filters...]
               [leftFilterGroup..., newFilterGroup, rightFilterGroups...]
             else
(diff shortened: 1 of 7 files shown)


modules/view/main/index.coffee
```diff
@@ -77,7 +77,7 @@ class View
 
       newNodeSet = enterSet.map (d, i) =>
         datum = d.datum
-        element = @new.call @rootSelection, d.datum, i
+        element = @new.call @rootSelection, datum, i, data.indexOf(datum)
```
c-frater: The advanced search needs this 'true index' which is difficult to retrieve outside the view
This new index differs from 'i' because 'i' is the index of the element in the list of elements being added whereas this is the index of the item in the complete dataset.

modules/toggle/main/index.scss
```diff
@@ -43,7 +43,7 @@
   }
 
   &.hx-btn > .hx-toggle-box {
-    margin-left: 1em;
+    margin-left: 0.5em;
```
c-frater: These styles are to make the toggle smaller

modules/data-table/test/spec.coffee
```diff
@@ -1858,101 +2063,6 @@ describe 'data-table', ->
           setTimeout(f, 50)
 
 
-  describe 'url feed', ->
```
c-frater: Restructured the tests a little - this is now nested in the 'data feeds' section (where it was probably supposed to be from the off)

modules/data-table/main/theme.scss
```diff
@@ -51,9 +51,11 @@
     }
   }
 
-  .hx-data-table-footer {
+  .hx-data-table-control-panel-bottom,
+  .hx-data-table-control-panel-compact,
+  .hx-data-table-control-panel {
```
c-frater: There is no longer a concept of a 'footer' as it can appear above or below the table
There is the compact control panel that houses the 'menu' button and a paginator, the main control panel that contains everything and the bottom control panel that contains a paginator and a page size picker

modules/data-table/main/theme.scss
```diff
@@ -51,9 +51,11 @@
...
     background: $footer-background-col;
-    box-shadow: 1px 1px 1px 1px $footer-shadow-col;
+    box-shadow: 1px 1px 1px $footer-shadow-col;
```
c-frater: This was not consistent with the style for the compact rows - It looked nicer to change it this way as the box shadow is slightly smaller

modules/data-table/test/spec.coffee
```diff
@@ -266,19 +275,16 @@ describe 'data-table', ->
           hx.select(rows[2]).selectAll('td').selectAll('.hx-data-table-cell-key').text().should.eql(headers)
           hx.select(rows[2]).selectAll('td').selectAll('.hx-data-table-cell-value').text().should.eql(['Dan', '41', 'Builder'])
 
-        it 'should show the paginator block', ->
-          container.select('.hx-data-table-pagination-block-hidden').size().should.equal(0)
-
         it 'should not show the paginator', ->
-          container.select('.hx-data-table-multi-page').size().should.equal(0)
+          container.selectAll('.hx-data-table-paginator-visible').size().should.equal(0)
```
c-frater: I've standardised some classes - it was difficult to follow what was going on in some places and in others the structure has changed so the old classes didn't make sense.
 
In this case, there are now 3 paginators that are made visible by adding this class to them individually to make it easier to manage

modules/data-table/test/spec.coffee
```diff
@@ -266,19 +275,16 @@ describe 'data-table', ->
...
-          container.select('.hx-data-table-page-size').size().should.equal(1)
-          container.select('.hx-data-table-page-size').classed('hx-data-table-select-page-size').should.equal(false)
+          container.selectAll('.hx-data-table-page-size').size().should.equal(2)
+          container.selectAll('.hx-data-table-page-size-visible').size().should.equal(0)
 
         it 'should show the filter box', ->
-          container.select('.hx-data-table-filter-control').size().should.equal(1)
-          container.select('.hx-data-table-filter-control').classed('hx-data-table-filter-visible').should.equal(true)
+          container.select('.hx-data-table-filter').size().should.equal(1)
+          container.select('.hx-data-table-filter').classed('hx-data-table-filter-visible').should.equal(true)
```
c-frater: The `control` suffix didn't really make sense so I figured I'd change it as the rest of the structure has already changed quite a bit

modules/data-table/test/spec.coffee
```diff
@@ -465,7 +467,7 @@ describe 'data-table', ->
               done()
 
             container
-              .select('.hx-data-table-pagination-picker')
+              .select('.hx-data-table-paginator-picker')
```
c-frater: changed everywhere we use 'pagination' to 'paginator' because it was confusing having the two different variations for similar components

modules/data-table/test/spec.coffee
```diff
@@ -1558,6 +1560,114 @@ describe 'data-table', ->
           ])
           done()
 
+
+      describe 'advanced search', ->
```
c-frater: New tests for the advanced search inside the feed - Tests are still in progress for the _actual_ advanced search, this just deals with the data and making sure it is filtered using Distributed Normal Form

modules/data-table/main/index.coffee
```diff
@@ -59,100 +265,182 @@ class DataTable extends hx.EventEmitter
...
+          .add(advancedSearchButtons
+            .add(advancedSearchAddFilterButton)
+            .add(advancedSearchClearFilterButton))))
+      # Bottom control panel - shown in compact mode and when the search is at the top
+      .add(controlPanelBottom
+        .add(spacer())
+        .add(pageSizeBottom)
+        .add(paginationBottom))
+      # Add the loading div last - helps keep it on top of everything
+      .add(loadingDiv)
```
c-frater: After scrolling up and down through the constructor repeatedly to work out how everything was constructed, I converted everything to detached divs and added the structure in one place - I think it's much easier to work out this way

modules/data-table/main/index.coffee
```diff
@@ -59,100 +265,182 @@ class DataTable extends hx.EventEmitter
...
+      clearFiltersText: hx.userFacingText('dataTable', 'clearFilters')
+      anyColumnText: hx.userFacingText('dataTable', 'anyColumn')
 
-    resolvedOptions.pageSize = Math.min resolvedOptions.pageSize, 1000
+      advancedSearchText: hx.userFacingText('dataTable','advancedSearch')
+      advancedSearchPlaceholder: hx.userFacingText('dataTable', 'search')
+    }, options)
 
-    onInput = hx.debounce 200, => @filter(filterInput.value(), undefined, 'user')
+    resolvedOptions.pageSize = Math.min resolvedOptions.pageSize, 1000
```
c-frater: Because I've changed how the structure is built, it lets us define things in a better order.
Unfortunately, this seems to have messed up the diff a little...

modules/data-table/main/index.coffee
```diff
@@ -200,6 +488,10 @@ class DataTable extends hx.EventEmitter
   displayMode: option('displayMode')
   feed: option('feed')
   filter: option('filter')
+  advancedSearch: option('advancedSearch')
```
c-frater: advancedSearch is independent of filter - this lets us switch between the two in a much nicer fashion without complicating the filter code

modules/data-table/main/index.coffee
```diff
@@ -727,19 +1064,42 @@ class DataTable extends hx.EventEmitter
...
+    cellValueLookup: hx.identity
+    termFormatter: defaultTermFormatter
+
     #XXX: should this provide more information - like the column id being sorted on?
-    compare: (c1, c2) -> if c1 > c2 then 1 else -1
+    compare: hx.sort.compare
   }, options)
 
+  options.filter ?= (term, row) -> options.termFormatter(term, getRowSearchTerm(options.cellValueLookup, row))
+  options.advancedSearch ?= getAdvancedSearchFilter(options.cellValueLookup, options.termFormatter)
```
c-frater: I've updated the object feed to be more configurable - mainly because it could be difficult for people to define a custom advanced search function.
Additions are the cellValueLookup to allow cell values to be non-strings and termLookup to allow the term to be split or formatted in whatever way is required

modules/data-table/main/index.coffee
```diff
@@ -837,3 +1204,4 @@ hx.dataTable = (options) ->
 
 hx.dataTable.objectFeed = objectFeed
 hx.dataTable.urlFeed = urlFeed
+hx.dataTable.getAdvancedSearchFilter = getAdvancedSearchFilter
```
c-frater: I've exposed the function for getting the advanced search filter as it would be a useful thing to have when writing a custom feed. I'm not sure if it's required though.

modules/data-table/main/index.coffee
```diff
@@ -15,6 +21,203 @@ hx.userFacingText({
...
+  if options.columns isnt undefined and options.columns[id] isnt undefined and options.columns[id][name] isnt undefined
+    options.columns[id][name]
+  else
+    options[name]
+
+splitArray = (array, index) ->
+  left = if index is 0
+    []
+  else
+    array[0..index - 1]
```
jmsmyth: `array[0...index]`

jmsmyth: Actually, maybe this would be clearer without the newlines?

```
left = if index is 0 then [] else array[0...index]
right = if index is array.length - 1 then [] else array[index+1...array.length]
```

modules/data-table/main/index.coffee
```diff
@@ -15,6 +21,203 @@ hx.userFacingText({
...
+  right = if index is array.length - 1
+    []
+  else
+    array[index+1..array.length]
```
jmsmyth: Also may want `array[index+1...array.length]` here, otherwise you'll be indexing out of range (which will work since this gets converted into a slice, but is probably not what you want)

modules/data-table/main/index.coffee
c-frater: I agree, it does make more sense. I think there was some confusion when I was writing this as to what `..` meant vs `...` - It's cleaner to put it on single lines as well

modules/data-table/test/spec.coffee
```diff
@@ -266,19 +303,25 @@ describe 'data-table', ->
           hx.select(rows[2]).selectAll('td').selectAll('.hx-data-table-cell-key').text().should.eql(headers)
           hx.select(rows[2]).selectAll('td').selectAll('.hx-data-table-cell-value').text().should.eql(['Dan', '41', 'Builder'])
 
-        it 'should show the paginator block', ->
-          container.select('.hx-data-table-pagination-block-hidden').size().should.equal(0)
```
c-frater: The paginator block no longer exists because of the restructure required for the advanced search

modules/data-table/test/spec.coffee
```diff
@@ -295,7 +338,7 @@ describe 'data-table', ->
       it 'should add the allowHeaderWrap class to all columns when enabled by default', (done) ->
         tableOptions =
           allowHeaderWrap: true
-        testTable {tableOptions: tableOptions}, done, (container, dt, options, data) ->
+        testTable {tableOptions}, done, (container, dt, options, data) ->
```
c-frater: Make use of coffee destructuring

modules/data-table/test/spec.coffee
```diff
@@ -1858,99 +2111,564 @@ describe 'data-table', ->
...
+              cells: { 'name': 'Dan', 'age': 41, 'profession': 'Builder' }
+            },
+            {
+              id: 0, # hidden details can go here (not in the cells object)
+              cells: { 'name': 'Bob', 'age': 25, 'profession': 'Developer' }
+            }
+          ])
+          done()
+
+      it 'should use the provided filter function', ->
```
c-frater: New tests for object feeds start here - the ones above haven't changed but have been moved in the file

c-frater:

c-frater:
This also resolves #252 due to the re-structuring

PR #268, opened 2016-07-24
Fix hx.minBy and hx.maxBy

jmsmyth:
`hx.minBy` and `hx.maxBy` have an off-by-one error meaning the last value in the array won't be checked. This change fixes that.

Also, the handling for undefined values has been tweaked. Before `undefined` values would be ignored, but now they are passed to the handler to be checked - it is feasible that you might want to include `undefined` values in a `hx.minBy` or `hx.maxBy` call.


c-frater:

PR #269, opened 2016-07-24
Add argmin/argmax

jmsmyth:
Add functions for getting the position of the minimum/maximum values in an array (with an optional lookup function)

### DIFF
--- a/modules/util/main/index.coffee
+++ b/modules/util/main/index.coffee
@@ -89,6 +89,30 @@ hx.minBy = (values, f) ->
         min = v
     min
 
+hx.argmin = (values, f) ->
+  if not values? or values.length is 0 then return undefined
+
+  minIndex = 0
+  minValue = undefined
+  if f
+    minValue = f(values[0])
+    if values.length > 1
+      for i in [1...values.length] by 1
+        v = f(values[i])
+        if minValue is undefined or (v isnt undefined and v < minValue)
+          minValue = v
+          minIndex = i
+  else
+    minValue = values[0]
+    if values.length > 1
+      for i in [1...values.length] by 1
+        v = values[i]
+        if minValue is undefined or (v isnt undefined and v < minValue)
+          minValue = v
+          minIndex = i
+
+  if minValue is undefined then undefined else minIndex
+
 hx.max = (values) -> Math.max.apply(null, values?.filter(hx.defined))
 
 hx.maxBy = (values, f) ->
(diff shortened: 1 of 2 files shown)


modules/util/main/index.coffee
```diff
@@ -89,6 +89,30 @@ hx.minBy = (values, f) ->
...
+          minValue = v
+          minIndex = i
+  else
+    minValue = values[0]
+    if values.length > 1
+      for i in [1...values.length] by 1
+        v = values[i]
+        if minValue is undefined or (v isnt undefined and v < minValue)
+          minValue = v
+          minIndex = i
```
c-frater: You could get rid of duplication by using hx.identity as the default value of `f` here

c-frater:
Are these two functions not aliases for `hx.minBy` and `hx.maxBy`?

modules/util/test/spec.coffee
```diff
@@ -157,6 +165,14 @@ describe "Util", ->
     hx.maxBy([1, 5, 3, 4]).should.equal(5)
     hx.maxBy([5, 1, undefined, 5, 4], (d) -> -d).should.equal(1)
 
+  it 'argmax: should return the index of the largest defined value', ->
+    hx.argmax([1, 4, -5, 9]).should.equal(3)
+    hx.argmax([undefined, -5, 9]).should.equal(2)
+    hx.argmax([1, undefined, -5, 9]).should.equal(3)
+    should.not.exist(hx.argmax([undefined, undefined]))
+    should.not.exist(hx.argmax([]))
+    should.not.exist(hx.argmax())
```
c-frater: Tests using the lookup function?

jmsmyth: Good point. I'll add some

modules/util/main/index.coffee
jmsmyth: Could do, but I did it this way to avoid the extra function call when dealing with number arrays. I'll benchmark to see how much overhead that actually adds. Maybe hx.identity gets optimised away...

c-frater: I think in this case it would probably be better to go with less code over performance - although I don't think the difference will be that noticeable :)

c-frater:

georgesimms:
Are these good names ? Would `maxIndex` be clearer ?

georgesimms:
Is the behavior well defined in the case of duplicates ?

c-frater:
@georgesimms I figured the names were suitable as they refer to the mathematical theory - thinking about it though it might make more sense to call them `maxIndex` and `minIndex`, we could always add an alias to argmin/argmax if needed?

PR #276, opened 2016-07-29
Meter fixes

### LINKED ISSUE #270: Meter: should emit an event on render
As the render is not handled explicitly by the user (i.e. render is called on resize), rendering the meter should emit an event.

### LINKED ISSUE #271: Meter: 'of' text is not localised
The `of 100` text displayed on the meter is not localised using `hx.userFacingText`

### LINKED ISSUE #272: Meter: add formatters for current/total value
Currently the values are displayed as passed in.
In some cases, the values need to be passed in as floating point numbers to display the meter accurately whilst the text values need to be rounded, it is currently not possible to do this.

jonthans:
fixed several issues for the meter
resolves #270
resolves #271
resolves #272

### DIFF
--- a/modules/meter/main/index.coffee
+++ b/modules/meter/main/index.coffee
@@ -72,8 +72,7 @@ class Meter extends hx.EventEmitter
     else
       @_.data
 
-  render: (cause) ->
-    cause ?= 'api'
+  render: (cause = 'api') ->
     _ = @_
     options = @options
     data = _.data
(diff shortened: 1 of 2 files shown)


modules/meter/main/index.coffee
```diff
@@ -64,7 +72,8 @@ class Meter
     else
       @_.data
 
-  render: ->
+  render: (cause) ->
+    cause ?= 'api'
```
georgesimms: Default argument instead ?

PR #277, opened 2016-07-30
Plot redraw option

### LINKED ISSUE #152: Plot: add an option to disable automatic graph redraw when the containing element is resized

jonthans:
Added redraw option
resolves #152

### DIFF
--- a/modules/plot/main/coffee/Sparkline.coffee
+++ b/modules/plot/main/coffee/Sparkline.coffee
@@ -37,7 +37,7 @@ class Sparkline
     hx.components.clear(selector)
     hx.component.register(selector, this)
 
-    graph = new hx.Graph(selector)
+    graph = new hx.Graph(selector, options)
 
     if opts.type isnt 'bar' and opts.type isnt 'line'
       hx.consoleWarning('options.type can only be "line" or "bar", you supplied "' + opts.type + '"')
--- a/modules/plot/main/coffee/Sparkline.coffee
+++ b/modules/plot/main/coffee/Sparkline.coffee
@@ -37,7 +37,7 @@ class Sparkline
     hx.components.clear(selector)
     hx.component.register(selector, this)
 
-    graph = new hx.Graph(selector, options)
+    graph = new hx.Graph(selector, { redrawOnResize: options.redrawOnResize })
 
     if opts.type isnt 'bar' and opts.type isnt 'line'
       hx.consoleWarning('options.type can only be "line" or "bar", you supplied "' + opts.type + '"')
--- a/modules/plot/main/coffee/Sparkline.coffee
+++ b/modules/plot/main/coffee/Sparkline.coffee
@@ -70,6 +70,9 @@ class Sparkline
   fillColor: optionSetterGetter('fillColor')
   strokeColor: optionSetterGetter('strokeColor')
   labelRenderer: optionSetterGetter('labelRenderer')
+  redrawOnResize: (value) ->
+    @_.graph.redrawOnResize(value)
+    optionSetterGetter('redrawOnResize').apply(this, arguments)
 
   render: ->
     self = this
--- a/modules/plot/test/spec.coffee
+++ b/modules/plot/test/spec.coffee
@@ -110,7 +110,7 @@ describe "plot", ->
       fakeNodeEvent(selection.node(), 'resize')()
       renderSpy.should.not.have.been.called()
 
-    it 'should by default redraw on resize', ->
+    it 'should redraw on resize by default', ->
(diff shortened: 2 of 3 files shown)


modules/plot/test/spec.coffee
```diff
@@ -1,8 +1,12 @@
+# Used to mimic an event call for a node
+fakeNodeEvent = (node, eventName) ->
+  if node?
+    (e) -> hx.select.getHexagonElementDataObject(node).eventEmitter?.emit((if eventName? and isNaN(eventName) then eventName else 'click'), e)
```
georgesimms: Why is isNaN being used here ? Are you anticipating a number string here ?

Would it be clearer to pull out the hexagon event name to a variable?

modules/plot/test/spec.coffee
```diff
@@ -91,6 +95,37 @@ describe "plot", ->
       axis.x.scaleType().should.equal('discrete')
       axis.y.scaleType().should.equal('time')
 
+    it 'should not redraw on resize when told not to', ->
```
georgesimms: Might it be worth testing the other way around too ? Ie check it does redraw on resize by default.

Mainly as a precaution against errors in this very test.

modules/plot/test/spec.coffee
georgesimms: also I can envisage this being useful in multiple tests. Might it be worth having a "test helper" module ?

jonthans: That was my guess as well, that we could extract it to somewhere shared. I'm not sure how often it is already being used.
This is copied from the data-table spec (Charlie told me to do the same yesterday with the meter)

modules/plot/test/spec.coffee
jonthans: see next test further down :-)

jmsmyth:
I wonder if this option should also have a setter getter method for changing the option after the graph has been created. Probably fine to leave out for now, and add in later if the need arises

Other than that, I'd say this looks good to merge

modules/plot/test/spec.coffee
c-frater: We currently don't have a 'test helpers' file but I agree that one would be very useful - there are several things that we could put in there (e.g. emulating selecting a menu item etc)

modules/plot/test/spec.coffee
```diff
@@ -91,6 +95,37 @@ describe "plot", ->
...
+      axis.addSeries('line', data: [{ x: 0, y: 1 }])
+      graph.render()
+
+      renderSpy = chai.spy()
+      graph.on 'render', renderSpy
+      selection.style 'width', '400px'
+      fakeNodeEvent(selection.node(), 'resize')()
+      renderSpy.should.not.have.been.called()
+
+    it 'should by default redraw on resize', ->
```
c-frater: I'd suggest `should redraw on resize by default`

modules/plot/main/coffee/Sparkline.coffee
```diff
@@ -37,7 +37,7 @@ class Sparkline
     hx.components.clear(selector)
     hx.component.register(selector, this)
 
-    graph = new hx.Graph(selector)
+    graph = new hx.Graph(selector, options)
```
c-frater: I'm not sure what the additional options that are specific for sparkline will do if passed into the graph directly, I'd suggest adding an option for `graphOptions` and passing that through instead? That way it's a bit more explicit (we use this method for components added to Form Builder)

jonthans: I just found it weird that the Sparkline has no control over the Graph it is rendering at all... Why is Sparkline not extending Graph but just using it in the background anyway?

c-frater: The idea behind sparkline is that it is a graph for a specific purpose with _intentionally_ limited options that prevent you from making it non-sparkline. That being said, some of the options are useful so we could either allow all options to be set (and let people do what they want) or pass through only the options that wouldn't affect the sparkline-ness of the graph.
In it's current state, I wouldn't want to merge it as passing through the complete options object seems a bit haphazard and may introduce issues down the line as there are sparkline specific options being used

jonthans: Ok. So I'll take that out again for now (just leaving the redrawOnResize option)

jonthans:
@jmsmyth: I'm not sure why you'd want a getter/setter for that option. You should know whether you want the option or not at the point when you create the graph?

c-frater:
We have setter getters for all the other options - in terms of consistency it would make sense to have this as well :P

PR #278, opened 2016-08-01
added shared test-helpers file

jonthans:


build/main/karma.js
```diff
@@ -26,27 +26,33 @@ var flatten = require('flatten')
...
+        ]
+      })
+    })).then(function (files) {
+      return runKarma(flatten(files), destDir, phantomOnly)
     })
-  })).then(function (files) {
-    return runKarma(flatten(files), destDir, phantomOnly)
+  }).catch((e) => {
+    console.log(e)
+    process.exit(1)
```
c-frater: We probably don't need this catch anymore

build/main/test-helpers.js
```diff
@@ -0,0 +1,15 @@
+var testHelpers = (function () {
+  var fakeNodeEvent = function (node, eventName) {
+    if (node != null) {
+      return function (e) {
+        var eventEmitter = hx.select.getHexagonElementDataObject(node).eventEmitter
+        if (eventEmitter != null) {
+          // we're using fakeNodeEvent for mapping arrays, so the eventName might be a an index
```
c-frater: `a an` :P

modules/data-table/test/spec.coffee
```diff
@@ -405,7 +400,7 @@ describe 'data-table', ->
           dt.expandedRows().should.eql([])
 
           clickHandlers = container.select('.hx-sticky-table-header-left').select('tbody').selectAll('.hx-data-table-collapsible-toggle')
-            .nodes.map(fakeNodeEvent)
+            .nodes.map(testHelpers.fakeNodeEvent)
```
c-frater: This is what the comment is referring to in `test-helpers.js`:
`// we're using fakeNodeEvent for mapping arrays, so the eventName might be a an index`

The second argument in `map` is the index - we should probably change this to remove the need for checking isNaN:
```
.nodes.map((node) -> testHelpers.fakeNodeEvent(node)) 
```

PR #281, opened 2016-08-03
Notify: fix the flex styles

### LINKED ISSUE #280: hx.notify styles bug
Hexagon version: 1.4.2    

Notifications generated by hx.notify have incorrect layout, because the `flex-wrap: nowrap` style for `.hx-notification`

https://github.com/ocadotechnology/hexagonjs/blob/1.4.2/modules/notify/main/index.scss#L50

gets overridden by `.hx-horizontal`'s `flex-flow: row wrap;`.    

So the `.hx-notification` element ends up having `flex-wrap: wrap`, which causes icons and text to be on separate rows.

See the plunker:
https://plnkr.co/edit/TalyJxcI2TNhrKJIMgqa?p=preview

Screenshot:
hx-notify-layout-bug

P.S. on the same plunkr you can see that using 'cssclass' option makes sucess/error icons disappear, which I guess is another bug...

jmsmyth:
Fixes #280 

Whilst fixing the bug, I've removed the dependence on the layout module (so no longer uses the hx-group, hx-section, classes)

### DIFF
--- a/modules/notify/main/index.coffee
+++ b/modules/notify/main/index.coffee
@@ -2,19 +2,19 @@
 setupNotification = (notification, selection) ->
   if notification.options.icon? and notification.options.icon.length > 0
     selection.append('div')
-      .class('hx-notification-icon-container hx-section hx-fixed hx-no-margin')
+      .class('hx-notification-icon-container')
       .append('i')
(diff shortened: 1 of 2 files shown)


c-frater:
LGTM 👍

PR #282, opened 2016-08-04
add default value as true for sparkline 'redraw on resize'

### LINKED ISSUE #279: Plot: the redrawOnResize value is false by default for sparklines

c-frater:
When `redrawOnResize` was added, it was assumed that it would be defaulted to _true_ however as we pass it over in the options directly (`{ redrawOnResize: options.redrawOnResize }`) it was defaulting to `undefined`
Resolves #279


c-frater:
I need to add regression tests

PR #283, opened 2016-08-04
update tag input to correctly handle undefined values

### LINKED ISSUE #264: Tag Input: calling the add method with undefined adds an empty tag
e.g. 
```js
tagInput.items() // = []
tagInput.add(undefined)
tagInput.items() // = [""]
```

By extension, this also means that calling the `items` method with undefined also creates the same issue:
```js
tagInput.items() // = []
tagInput.items(undefined)
tagInput.items() // = [""]
```

The items method should only accept an array of items and the add method should not accept undefined.

Expected outcome:
```js
tagInput.add(undefined) // shows a warning (must be defined), does nothing
tagInput.items(undefined) // sets the items to [] - most likely the desired behaviour
tagInput.items("something") // shows a warning (must be an array), does nothing
```

c-frater:
resolves #264

### DIFF
--- a/modules/tag-input/main/index.coffee
+++ b/modules/tag-input/main/index.coffee
@@ -116,8 +116,14 @@ class TagInput extends hx.EventEmitter
   add: (name, cssclass) ->
     if hx.isArray(name)
       addTag(this, n, cssclass) for n in name
-    else
+    else if name
       addTag(this, name, cssclass)
+    else
+      hx.consoleWarning(
+        'TagInput.add was passed the wrong argument type',
+        'TagInput.add accepts an array or string argument, you supplied:',
+        name
+      )
     @input.value('')
     if @options.draggable then @_.dragContainer.setup()
     this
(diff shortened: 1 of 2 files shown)


modules/tag-input/main/index.coffee
```diff
@@ -116,8 +116,14 @@ class TagInput extends hx.EventEmitter
...
       addTag(this, n, cssclass) for n in name
-    else
+    else if name
       addTag(this, name, cssclass)
+    else
+      hx.consoleWarning(
+        'TagInput.add was passed the wrong argument type',
+        'TagInput.add accepts an array or string argument, you supplied:',
+        name
+      )
```
georgesimms: Can this use the "items" method ?

c-frater: It could but I don't think it should.
Add is for adding a single item to the list whereas items is for adding multiple items to the list.
We could add the ability to add an array of tags but I'm not sure why you'd ever need to do that as opposed to just using `items` to set the list?

PR #284, opened 2016-08-04
prevent meter rendering when the height/width are 0

### LINKED ISSUE #170: Meter: when the width or height of the container is 0, an exception is thrown on render
The error looks something like this

```
Invalid value for <path> attribute d="M0,0LNaN,NaNLNaN,NaNLNaN,NaNLNaN,NaNz"`
```

The fix is to check that size is non-zero in the meter `render` method

c-frater:
add redrawOnResize to meter (mainly to make testing easier)
fix issue where redrawOnResize was not dynamic in graphs

resolves #170

### DIFF
--- a/modules/meter/main/index.coffee
+++ b/modules/meter/main/index.coffee
@@ -33,13 +33,15 @@ class Meter extends hx.EventEmitter
       trackerBackgroundCol: hx.color(hx.theme.palette.contrastCol).alpha(0.05).toString()
       markerCol: hx.theme.palette.contrastCol
       valueFormatter: (value, isTotal) -> if isTotal then "#{hx.userFacingText('meter', 'of')} #{value}" else value
+      redrawOnResize: true
     }, options)
 
     randomId = hx.randomId()
 
     _.selection = selection = hx.select(selector)
       .classed('hx-meter', true)
-      .on 'resize', => @render('user')
+      .on 'resize', =>
+        @render('user') if @options.redrawOnResize
 
     _.container = container = selection.append('div').class('hx-meter-container')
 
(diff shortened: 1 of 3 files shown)


modules/meter/main/index.coffee
```diff
@@ -33,13 +33,15 @@ class Meter extends hx.EventEmitter
       trackerBackgroundCol: hx.color(hx.theme.palette.contrastCol).alpha(0.05).toString()
       markerCol: hx.theme.palette.contrastCol
       valueFormatter: (value, isTotal) -> if isTotal then "#{hx.userFacingText('meter', 'of')} #{value}" else value
+      redrawOnResize: true
```
c-frater: I think this probably warrants making it a `Minor` release but I also can't decide if it could go in with a `Patch`?

modules/meter/main/index.coffee
```diff
@@ -89,89 +91,90 @@ class Meter extends hx.EventEmitter
 
     size = Math.min(selection.width(), selection.height() * 2)
 
-    # 14 is the standard font size, 150 is the assumed size, and size is the actual
-    # size of the container. This scales the fonts appropriately
-    container.style('font-size', Math.round(14 / 150 * (size / 2)) + 'px')
-    textOffset = 10 * (size / 2) /  150
+    if size > 0
```
c-frater: If size is 0 do nothing and return the meter. Currently it doesn't emit the `render` event when the size is 0, is this the right behaviour?

modules/meter/test/spec.coffee
```diff
@@ -130,3 +160,30 @@ describe 'hx-meter', ->
 
     selection.select '.hx-meter-total'
       .text().should.equal 'total'
+
+  it 'should not render when the container size is 0', ->
```
c-frater: New test

modules/plot/main/coffee/Graph.coffee
```diff
@@ -35,7 +35,9 @@ class Graph extends hx.EventEmitter
 
     selection = hx.select(@selector)
 
-    if @_.options.redrawOnResize then selection.on 'resize', 'hx.plot', => @render()
+    selection.on 'resize', 'hx.plot', =>
+      @render() if @_.options.redrawOnResize
```
c-frater: This fixes the issue where setting up graph with `redrawOnResize` and then changing the value using the setter/getter doesn't actually _do_ anything because the 'resize' event is created in the constructor

modules/meter/main/index.coffee
```diff
@@ -89,89 +91,90 @@ class Meter extends hx.EventEmitter
...
-        y: y + radius * Math.sin((i/100 + 1)*Math.PI)
-
-      path = hx.plot.svgCurve(points, false)
-
-      _.markerTextCurve.attr('d', path)
-
-      if markerProgress < 0.5
-        _.markerText
-          .attr('startOffset', (markerProgress*100+1) + '%')
-          .attr('text-anchor', 'start')
```
georgesimms: Are these changes just indentation ? If so they are quite hard to read.

c-frater: Yes - the only change is that the render function (after `if size > 0`) has been indented to be wrapped by the if statement.
Github just hasn't worked out the diff very nicely

PR #285, opened 2016-08-04
fix button alignment inside forms

### LINKED ISSUE #274: Form: button margins are causing misalignment
The left margin is the culprit but adding something to remove the margin on all sides would also resolve the slight vertical misalignment caused by buttons taking up too much vertical space.

screen shot 2016-07-28 at 15 46 56

Using `hx-no-margin` on buttons inside the form will have the same effect for now.

This also affects Form Builder

c-frater:
resolves #274

### DIFF
--- a/modules/form/main/index.scss
+++ b/modules/form/main/index.scss
@@ -11,15 +11,24 @@
       display: table-cell;
     }
 
-    > *:not(input) {
+    > *:not(input):not(select):not(textarea):not(button):not(.hx-btn) {
       padding-top: 8px;
     }
 
     // Padding top doesn't look right for inputs and select and the other margins make the boxes difficult to align
     > input,
     > select,
-    > textarea {
+    > textarea,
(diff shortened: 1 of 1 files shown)


modules/form/main/index.scss
```diff
@@ -11,17 +11,26 @@
       display: table-cell;
     }
 
-    > *:not(input) {
+    > *:not(input):not(select):not(textarea):not(button):not(.hx-btn) {
```
georgesimms: Would {code}:not(input, select, textarea, button, .hx-btn){code} work ? It might be more readable ?

c-frater: You know, I've never actually thought to try doing multiple selectors inside a `:not` - I've no idea if it works.

c-frater: After some fiddling, it doesn't seem to like multiple selectors

PR #287, opened 2016-08-04
[WIP] work out a way to stop the intermittent test failures

c-frater:
This is to resolve an _intermittent_ issue with travis where the tests pass but the build fails when uploading the coverage.
This is more to test the theory than anything because it's annoying that 1/5 builds _occasionally_ fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ node_js:
 - '4'
 - '0.12'
 - '0.11'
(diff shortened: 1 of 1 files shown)


georgesimms:
I'm not sure I fully understand why this change works.

c-frater:
It _sort of_ works - It makes the `upload coverage` step fail less (tested it on multiple builds) but it still fails occasionally
It's an ongoing change

PR #291, opened 2016-08-09
#177 add increment on hold to number pickers

### LINKED ISSUE #177: Time Picker / Number Picker: Press and hold to increment/decrement
It should be easy enough to add press and hold detection to number pickers that let you hold down the button to increment/decrement.
This was raised as a feature request for time pickers when incrementing/decrementing the minutes.

UX Notes:
- There should be a slight delay between the mousedown and the start of the increment/decrement to account for long presses (around 200ms?)
- The increment speed should be reasonable enough that you can stop it on the intended number 90-100% of the time but not so slow that it's inconvenient (around 25ms?)
- It should be easier than clicking the button multiple times

c-frater:
resolves #177

### DIFF
--- a/modules/number-picker/main/index.coffee
+++ b/modules/number-picker/main/index.coffee
@@ -7,22 +7,32 @@ checkValue = (numberPicker, context) ->
   if value isnt oldValue then context.value(value)
 
 addHoldHandler = (incrementOnHold, incrementDelay, selection, incrementFn) ->
-  holdStart = undefined
-  holdTimer = undefined
-  incrementHold = undefined
   if incrementOnHold
-    selection.on 'pointerdown', 'hx.number-picker', ->
+    holdStart = undefined
+    holdTimeout = undefined
+    incrementInterval = undefined
+
+    clearTimers = ->
+      clearTimeout(holdTimeout)
+      clearInterval(incrementInterval)
+
+    selection.on 'pointerdown', 'hx.number-picker', (e) ->
       holdStart = Date.now()
-      fn = -> incrementHold = setInterval((-> incrementFn()), incrementDelay)
-      holdTimer = setTimeout(fn, 200)
+      document.activeElement.blur()
+      e.event.preventDefault()
+      fn = ->
+        incrementInterval = setInterval((-> incrementFn()), incrementDelay)
+      holdTimeout = setTimeout(fn, 200)
 
     selection.on 'pointerup', 'hx.number-picker', ->
-      if Date.now() - holdStart < 200
+      clearTimers()
+      if (Date.now() - holdStart) < 200
         incrementFn()
-      clearTimeout(holdTimer)
-      clearInterval(incrementHold)
+
+    selection.on 'pointerleave', 'hx.number-picker', clearTimers
   else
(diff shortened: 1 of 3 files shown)


c-frater:
Need to add tests

modules/number-picker/main/index.coffee
```diff
@@ -6,6 +6,24 @@ checkValue = (numberPicker, context) ->
...
 
+addHoldHandler = (incrementOnHold, incrementDelay, selection, incrementFn) ->
+  holdStart = undefined
+  holdTimer = undefined
+  incrementHold = undefined
+  if incrementOnHold
+    selection.on 'pointerdown', 'hx.number-picker', ->
+      holdStart = Date.now()
+      fn = -> incrementHold = setInterval((-> incrementFn()), incrementDelay)
+      holdTimer = setTimeout(fn, 200)
```
jonthans: Does this conflict with press-and-hold on touch devices?

jonthans:
Can this get tested as well?

modules/number-picker/main/index.coffee
```diff
@@ -6,6 +6,34 @@ checkValue = (numberPicker, context) ->
...
+    holdTimeout = undefined
+    incrementInterval = undefined
+
+    clearTimers = ->
+      clearTimeout(holdTimeout)
+      clearInterval(incrementInterval)
+
+    selection.on 'pointerdown', 'hx.number-picker', (e) ->
+      holdStart = Date.now()
+      document.activeElement.blur()
```
c-frater: Forcing blur here makes it possible to type in the input and then increment/decrement without having to click outside of the number picker first.

modules/number-picker/main/index.coffee
```diff
@@ -6,6 +6,34 @@ checkValue = (numberPicker, context) ->
...
+      fn = ->
+        incrementInterval = setInterval((-> incrementFn()), incrementDelay)
+      holdTimeout = setTimeout(fn, 200)
+
+    selection.on 'pointerup', 'hx.number-picker', ->
+      clearTimers()
+      if (Date.now() - holdStart) < 200
+        incrementFn()
+
+    selection.on 'pointerleave', 'hx.number-picker', clearTimers
```
c-frater: This resolves an issue with desktops when moving the mouse out of selection and releasing outside.
This is a safer way of doing it than adding listeners to the body

modules/number-picker/main/index.coffee
```diff
@@ -16,31 +44,30 @@ class NumberPicker extends hx.EventEmitter
...
-    @selectInput = select.append('input')
+    @selectInput = selection.append('input')
     @selectInput.attr('type', 'number')
     @selectInput.on 'blur', 'hx.number-picker', =>
-      if not @selectInput.attr('readonly')?
-        checkValue(this, @selectInput)
-        @selectInput.attr('data-value', @selectInput.value())
       @emit 'input-change', {value: @value()}
-      @emit 'change', {value: @value()}
+      @value(undefined, @selectInput.value())
```
c-frater: Value should be the only way to set the value

PR #293, opened 2016-08-10
#171: Util: Add doctype validation

### LINKED ISSUE #171: Add console warning when not using <!DOCTYPE html>
Strange things happen when you do not use `<!DOCTYPE html>`. In particular, tag inputs page oddly and titlebar icons float above where they belong. Of course, the user should remember to do this. But if there were a console warning here, the user would find the root of the problem more quickly.

georgesimms:
Resolves #171

### DIFF
--- a/modules/util/main/index.coffee
+++ b/modules/util/main/index.coffee
@@ -10,6 +10,12 @@ hx.consoleWarning = (heading, messages...) ->
   console.warn.apply(console, [heading].concat(messages))
   console.trace('Stack Trace')
 
+validateDoctype = (doctype) ->
+  doctype? and doctype.name is 'html' and not (doctype.publicId or doctype.systemId)
+
+unless validateDoctype document.doctype
+  hx.consoleWarning 'Expected doctype: <!DOCTYPE html>, actual doctype: ', document.doctype, ' so you may have CSS problems.'
+
 # consistent string to int hashing
 hx.hash = (str, max) ->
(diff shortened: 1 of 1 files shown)


c-frater:
From discussion, this is probably something to wait for #265 to be completed - This would allow us to create a dedicated file for 'utils' that don't belong in the utils module.

The main reasoning for this is that currently, this code doesn't _belong_ anywhere. 
A missing or incorrect doctype would mainly causes issues with rendering CSS and not issues with JS. Having it in this module makes it possible to not use any modules containing styles but still see the warning which probably isn't sensible (particularly if you are intentionally using an older doctype and just want some of hexagon's Utils).
The alternate solution of including the file in the `base` module is probably the current most sensible but the current base module doesn't include any script (and probably _shouldn't_) so I think we'll leave this for now. 
I think that sums up the conversation we had @jmsmyth @georgesimms?

modules/util/main/index.coffee
```diff
@@ -10,6 +10,12 @@ hx.consoleWarning = (heading, messages...) ->
   console.warn.apply(console, [heading].concat(messages))
   console.trace('Stack Trace')
 
+validateDoctype = (doctype) ->
+  doctype? and doctype.name is 'html' and not (doctype.publicId or doctype.systemId)
```
jonthans: or `doctype?.name is 'html'`

PR #294, opened 2016-08-10
#180: Make dropdowns reusable

### LINKED ISSUE #180: Dropdown: improve reusability of dropdowns

georgesimms:
resolves #180

### DIFF
--- a/modules/dropdown/main/index.coffee
+++ b/modules/dropdown/main/index.coffee
@@ -162,7 +162,10 @@ class Dropdown extends hx.EventEmitter
   show: (cb) ->
     _ = @_
 
-    if not _.visible
+    if _.visible
+      @render()
+      cb?()
+    else
       _.visible = true
 
       _.dropdown = hx.select(hx._.dropdown.attachToSelector).append('div').attr('class', 'hx-dropdown')
@@ -217,7 +220,7 @@ class Dropdown extends hx.EventEmitter
         .morph()
           .with('fadein', 150)
           .and('expandv', 150)
-          .and =>
+          .and ->
             _.dropdown.animate().style('top', y + 'px', 150)
           .then =>
             if _.useScroll and _.dropdown?
--- a/modules/dropdown/main/index.coffee
+++ b/modules/dropdown/main/index.coffee
@@ -131,13 +131,12 @@ class Dropdown extends hx.EventEmitter
   dropdownContent: (dropdownContent) ->
     if arguments.length
       setupDropdown = dropdownContentToSetupDropdown dropdownContent
-      @_.dropdownContent = dropdownContent
       @_ = hx.shallowMerge @_, {
         setupDropdown,
         dropdownContent
       }
-      debugger
       @render()
+      this
     else
       @_.dropdownContent
 
@@ -157,6 +156,8 @@ class Dropdown extends hx.EventEmitter
 
   render: ->
     @_.setupDropdown @_.dropdown.node()
+    @emit 'render'
+    this
 
   show: (cb) ->
     _ = @_
--- a/modules/dropdown/main/index.coffee
+++ b/modules/dropdown/main/index.coffee
@@ -59,6 +59,16 @@ calculateDropdownPosition = (alignments, selectionRect, dropdownRect, windowRect
     direction: direction
   }
 
+dropdownContentToSetupDropdown = (dropdownContent) ->
+  setupDropdown = switch
+    when hx.isString dropdownContent
+      (node) -> hx.select(node).html(dropdownContent)
+    when hx.isFunction dropdownContent
+      dropdownContent
+    else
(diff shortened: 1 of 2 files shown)


modules/dropdown/main/index.coffee
```diff
@@ -103,8 +105,11 @@ class Dropdown extends hx.EventEmitter
 
     selection = hx.select(selector)
 
+    setupDropdown = dropdownContentToSetupDropdown dropdownContent
+
     @_ = {
       setupDropdown: setupDropdown,
+      dropdownContent: dropdownContent,
```
georgesimms: so we can create a getter for it

modules/dropdown/main/index.coffee
```diff
@@ -193,7 +216,7 @@ class Dropdown extends hx.EventEmitter
         .morph()
           .with('fadein', 150)
           .and('expandv', 150)
-          .and =>
+          .and ->
```
georgesimms: remove superfluous fat arrow

modules/dropdown/main/index.coffee
```diff
@@ -135,11 +140,29 @@ class Dropdown extends hx.EventEmitter
...
-    _ = @_
 
-    if not _.visible
-      _.visible = true
+    if not @_.visible
+      @_.visible = true
+      @render cb
+    this
+
+  render: (cb) ->
```
georgesimms: most of the render method does what the show method previously did.

The render method does nothing if the dropdown is not visible (in particular, it will never call the callback)

c-frater:
The `showstart` and `showend` events shouldn't be handled in `render` - we can use the render CB as a 'passthrough' to allow it to call the showend from show

modules/dropdown/main/index.coffee
```diff
@@ -59,6 +59,16 @@ calculateDropdownPosition = (alignments, selectionRect, dropdownRect, windowRect
...
 
+dropdownContentToSetupDropdown = (dropdownContent) ->
+  setupDropdown = switch
+    when hx.isString dropdownContent
+      (node) -> hx.select(node).html(dropdownContent)
+    when hx.isFunction dropdownContent
+      dropdownContent
+    else
+      hx.consoleWarning 'dropdown: dropdownContent is not a valid type. dropdownContent: ', dropdownContent
+      -> undefined
```
c-frater: +1 for pulling this out

modules/dropdown/main/index.coffee
```diff
@@ -135,11 +140,29 @@ class Dropdown extends hx.EventEmitter
...
     this
 
+  dropdownContent: (dropdownContent, cb) ->
+    if arguments.length
+      setupDropdown = dropdownContentToSetupDropdown dropdownContent
+      @_.dropdownContent = dropdownContent
+      @_ = hx.shallowMerge @_, {
+        setupDropdown,
+        dropdownContent
+      }
```
c-frater: Unless I've misunderstood why we're doing this, you're essentially setting dropdown content twice - we could remove the `shallowMerge` and just set `@_.setupDropdown` if `setupDropdown` is defined?

Also, if you set `setupDropdown` to be invalid, should we still attempt to render?

modules/dropdown/main/index.coffee
```diff
@@ -135,11 +140,29 @@ class Dropdown extends hx.EventEmitter
...
+
   show: (cb) ->
-    _ = @_
 
-    if not _.visible
-      _.visible = true
+    if not @_.visible
+      @_.visible = true
+      @render cb
+    this
```
c-frater: As render also checks for `@_.visible` could we just call `@render cb` outside the if statement and remove the explict `this` return?

c-frater:
Tests for new features?

modules/dropdown/main/index.coffee
```diff
@@ -123,6 +128,20 @@ class Dropdown extends hx.EventEmitter
       selection.on('mouseover', 'hx.dropdown', onmouseover)
       selection.on('mouseout', 'hx.dropdown', onmouseout)
 
+  dropdownContent: (dropdownContent) ->
+    if arguments.length
+      setupDropdown = dropdownContentToSetupDropdown dropdownContent
+      @_ = hx.shallowMerge @_, {
```
jonthans: Could this rather set the fields directly on `@_` rather than creating a new copy of it and reassigning that?

PR #295, opened 2016-08-12
#194 add autocompletion to tag-input

### LINKED ISSUE #194: Tag Input: add autocomplete support

jonthans:
resolves #194


demo/index.html
```diff
@@ -204,6 +204,16 @@
...
 
+      <div class="hx-section exampleSection">
+        <div class="example">
+          <h3>Autocomplete Picker</h3>
+          <div id="autocomplete-picker"></div>
+          <script>
+            new hx.AutocompletePicker('#autocomplete-picker', ['Dave', 'Jane', 'Bob', 'Dan', 'Steve', 'Alejandro'])
+          </script>
+        </div>
+      </div>
```
jonthans: I thought this was missing on the demo page so I added it. I guess we should do that for other things that are missing as well at some point?

demo/index.html
```diff
@@ -384,6 +394,18 @@
...
+        <div class="example">
+          <h3>Autocompleted Tag Input</h3>
+          <div id="autocomplete-tag-input"></div>
+          <script>
+            new hx.TagInput('#autocomplete-tag-input', {
+              autocompleteData: ['Jane', 'Bob', 'Dan', 'Steve', 'Alejandro']
+            })
+          </script>
+        </div>
+      </div>
```
jonthans: I found it particularly helpful to play around with this while developing, so I'd vote for adding this for every new module/feature

modules/autocomplete-picker/main/index.coffee
```diff
@@ -224,4 +224,4 @@ hx.autocompletePicker = (items, options) ->
   new AutocompletePicker(selection.node(), items, options)
   selection
 
-hx.AutocompletePicker = AutocompletePicker
\ No newline at end of file
+hx.AutocompletePicker = AutocompletePicker
```
jonthans: These new-lines at the end of a file are IntelliJ's default addition, which I don't think is a bad idea

c-frater:
LGTM 👍

PR #296, opened 2016-08-17
#89 remove `atEdge` from the findLabel util function

### LINKED ISSUE #89: Plot: labels are not shown for the first/last values for graphs
The following sparkline has 3 points but the label is only shown for the first point:
```
function random (i) {
  return {
    x: i,
    y: Math.random()
  }
}

var container = hx.select('body')
  .append('div')
  .style('width', '500px')
  .style('height', '50px')
  .style('margin', '0 auto')

sp = new hx.Sparkline(container.node(), {data: hx.range(3).map(random)})
sp.render()
```

c-frater:
resolves #89

### DIFF
--- a/modules/plot/main/coffee/Utils.coffee
+++ b/modules/plot/main/coffee/Utils.coffee
@@ -204,7 +204,6 @@ search = (array, find, lookup) ->
 findLabel = (array, find, interpolate, interpolateValues) ->
   i = search(array, find, (d) -> d.x)
   if i > -1
-    atEdge = i is 0 or i is array.length - 1
     if interpolate
       closest = array[i]
       dist = find - closest.x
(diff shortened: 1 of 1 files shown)


modules/plot/main/coffee/Utils.coffee
```diff
@@ -204,7 +204,6 @@ search = (array, find, lookup) ->
 findLabel = (array, find, interpolate, interpolateValues) ->
   i = search(array, find, (d) -> d.x)
   if i > -1
-    atEdge = i is 0 or i is array.length - 1
```
c-frater: I'm not sure why we did this, I can't think of a reason for it other than something unrelated to `LineSeries`

PR #298, opened 2016-08-29
added disabled colour

### LINKED ISSUE #195: Palette: add 'disabled' as a context
The current palette functions have an array of contexts to use for text/background/context which currently do not include `disabled` which, for most modules (including palette with `hx-background-disabled`), is a valid class. 

Should this be added as a context in the array of valid contexts?

Example:
```js
hx.palette.context(element, 'disabled')
hx.palette.backgroundContext(element, 'disabled')
```

jonthans:
resolves #195


jmsmyth:
This will be added in 1.7.0

Issue #306, opened 2016-09-15
Data-Table: Issue with page-selector for infinite feeds


jonthans:
If filteredCount is first not present but becomes present later on (i.e. the feed reached its end) then the pageSelector is not enabled

jmsmyth:
Fixed in 1.7.0

PR #307, opened 2016-09-15
fixed #305 and #306

### LINKED ISSUE #305: Data-Table: Issue with page-size-selector for infinite feeds
When filteredCount is undefined the pageSizeSelector is always visible but has no values irrespective of the pageSizeOptions.

jonthans:
resolves #305 and #306

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -831,6 +831,43 @@
               compact: false,
               pageSizeOptions: [1, 10, 50],
               pageSize: 2,
+              feed: hx.dataTable.objectFeed({
+                headers: [
+                    {name: 'Name', id: 'name', groups: ['Child Group 1', 'Parent Group']},
+                    {name: 'Age', id: 'age', groups: ['Child Group 1', 'Parent Group']},
+                    {name: 'Profession', id: 'profession', groups: ['Child Group 2', 'Parent Group']}
+                ],
+                rows: [
+                    {
+                        id: 0, // hidden details can go here (not in the cells object)
+                        cells: {'name': 'Bob', 'age': 25, 'profession': 'Developer'}
+                    },
+                    {
+                        id: 1,
+                        cells: {'name': 'Jan', 'age': 41, 'profession': 'Artist'}
+                    },
+                    {
+                        id: 2,
+                        cells: {'name': 'Dan', 'age': 41, 'profession': 'Builder'}
+                    }
+                ]
+              })
+            }).render()
+          </script>
+        </div>
+      </div>
+
+      <!-- grouped header data table -->
+      <div class="hx-section exampleSection">
+        <div class="example">
+          <h3>Lazy Data Table</h3>
+          <div id="lazy-data-table"></div>
+          <script>
+            var finalSize = undefined
+            new hx.DataTable('#lazy-data-table', {
+              compact: false,
+              pageSizeOptions: [1, 2, 3, 10, 50],
+              pageSize: 2,
               feed: {
                 headers: function (cb) {
                   cb([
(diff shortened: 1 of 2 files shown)


demo/index.html
```diff
@@ -831,27 +831,35 @@
...
-                    id: 1,
-                    cells: { 'name': 'Jan', 'age': 41, 'profession': 'Artist' }
-                  },
-                  {
-                    id: 2,
-                    cells: { 'name': 'Dan', 'age': 41, 'profession': 'Builder' }
-                  }
-                ]
-              })
+              feed: {
```
jonthans: updated the example to use a "lazy feed"

modules/data-table/main/index.coffee
```diff
@@ -663,8 +663,6 @@ class DataTable extends hx.EventEmitter
           start = undefined
           end = undefined
 
-        selection.classed('hx-data-table-infinite', totalCount is undefined)
```
jonthans: allow for the table to change whether it is still considered infinite or not by checking for that at a later point

modules/data-table/main/index.coffee
```diff
@@ -746,22 +746,21 @@ class DataTable extends hx.EventEmitter
               @_.sortColPicker.value({value: currentSort.column + currentSort.direction})
 
           # populate the page size picker if there are options set
-          if filteredCount isnt undefined and filteredCount > 0
```
jonthans: remove this if as the filteredCount seems unrelated to the block it is surrounding.
The problem with this is that it makes the pageSizeOptions always get rendered without any items

modules/data-table/main/index.coffee
```diff
@@ -677,6 +675,8 @@ class DataTable extends hx.EventEmitter
         feed.rows range, ({rows, filteredCount}) =>
           if options.displayMode is 'paginate'
             multiPage = false
+            selection.classed('hx-data-table-infinite', filteredCount is undefined)
```
jonthans: This is to prevent a seemingly orphaned "of <n>" shown without the pageSelector next to it.

jonthans:
After this change the totalCount only gets used when the user selects rows. Maybe that means that the check for totalCount can be moved down into the @updateSelected method and the option on the feed can be made optional

demo/index.html
jmsmyth: Could we also keep the object feed example in as well? - so have both this one and the old version, just to make any manual testing a bit easier

demo/index.html
```diff
@@ -831,27 +831,35 @@
...
+                      {
+                        id: 1,
+                        cells: { 'name': 'Jan', 'age': 41, 'profession': 'Artist' }
+                      },
+                      {
+                        id: 2,
+                        cells: { 'name': 'Dan', 'age': 41, 'profession': 'Builder' }
+                      }
+                    ].slice(range.start, range.end + 1),
+                    filteredCount: range.end >= 3 ? 3 : undefined
```
jmsmyth: `range.end >= 3 ? 3 : undefined` should be `range.end >= 2 ? 3 : undefined`, otherwise the pagination breaks for the example

jonthans: Well, the idea is that the table is able to see that the feed has ended and from thereon displays the final size. I have updated the example to hold on to the fact that the size is known as soon as `range.end >= 3` so that when you go back to only showing the first 2 rows you will see `1-2  of 3` in the paginator

jmsmyth: Just to clarify, the problem here is that when the page size cleanly divides the filtered count you can click the next button even when no data exists for the next page:

data-table-pagination-example

The same happens if the page size is set to 3.

I think this is simply because range.end uses 0 based indexing, so `range.end >= 3` is actually checking if we've gone beyond the end of the 'stream' by 1, rather than checking if its at end. So for this example checking `range.end >=2` would resolve the problem. (If I've misunderstood how it works, please correct me :-) )

This would be the same in the iterator/stream situation, where you could check something like `iterator.hasNext()` before deciding to return a value for `filteredCount` to avoid the empty page problem.

jonthans: gotcha :+1:

jmsmyth:
This will be released in 1.7.0

PR #314, opened 2016-09-30
#29: add lazy loading to trees

### LINKED ISSUE #29: Tree: add the option to build trees lazily when building from JSON

c-frater:
resolves #29 

I used this for benchmarking:
```js
let count = 0
const createItemTree = (levels, level = 0) => {
  return hx.range(10).map(int => {
    let children
    if (level < levels) {
      children = createItemTree(levels, level + 1)
    }
    count += 1
    return {
      name: 'Node',
      children: children || []
    }
  })
}
const itemTree = createItemTree(4)
console.log('countNodes', count)

start = new Date()
hx.select('body')
.add(hx.tree({
  lazy: true,
  items: itemTree
}))

hx.select('body')
  .add('br')

console.log('first', Date.now() - start.getTime())
start = new Date()
hx.select('body')
.add(hx.tree({
  lazy: false,
  items: itemTree
}))
console.log('second', Date.now() - start.getTime())
```

It's not the cleanest way of doing it but it detects the time it blocks execution of other JS.
```
countNodes 111110
first 312
second 22350
```

Double click is still enabled for lazy loading trees so if you have a lot of nested content in one node you can still 'hang' the UI - I'm looking at optimising the rendering so it only updates the DOM at the last step but in it's current state this resolves the issue :)

### DIFF
--- a/modules/tree/main/index.coffee
+++ b/modules/tree/main/index.coffee
@@ -1,5 +1,35 @@
+createNodeView = (node, renderer, lazy) ->
+  node.view('.hx-tree-node')
+    .enter (d) -> @append(createTreeNode(d, renderer, lazy)).node()
+
+createChildren = (children, renderer, lazy) ->
+  children.map (child) -> createTreeNode(child, renderer, lazy)
+
+createTreeNode = (data, renderer, lazy) ->
+  treeNode = hx.detached('div').class('hx-tree-node')
+
+  nodeContent = hx.detached('div').attr('class', 'hx-tree-node-content')
+  renderer(nodeContent.node(), data)
+  if data.children? and data.children.length > 0
+    childContainer = hx.detached('div').class('hx-tree-node-children').style('display', 'none')
+    if lazy
+      childView = createNodeView(childContainer, renderer, lazy)
+      hx.component.register(treeNode.node(), {
+        renderChildren: -> childView.apply(data.children)
+      })
+    else
+      childContainer.append(createChildren(data.children, renderer, lazy))
+
+    treeNode.add(hx.detached('div').class('hx-tree-node-parent')
+        .add(nodeContent))
+      .add(childContainer)
+  else
+    treeNode.add(nodeContent)
+
+formatChildren = (tree, children, animate) -> children.forEach((d) => format(tree, d.node(), animate))
+
 formatIcon = (node, iconElement, animate) =>
-  children = hx.select(node).select('.hx-tree-node-children')
+  children = hx.select(node).shallowSelect('.hx-tree-node-children')
 
   if not children.selectAll('.hx-tree-node').empty()
     open = children.style('display') == 'block'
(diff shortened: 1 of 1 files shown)


modules/tree/main/index.coffee
```diff
@@ -82,22 +123,20 @@ class Tree
     @options = hx.merge.defined {
       hideDisabledButtons: false
       animate: true
-      renderer: (elem, data) -> hx.select(elem).html(data)
+      renderer: (elem, data) -> hx.select(elem).html(data.name || data)
```
c-frater: I figured adding this to the default renderer made sense as there's pretty much no point in having a tree that only has one level and to get more levels you need an object...

modules/tree/main/index.coffee
```diff
@@ -176,9 +185,7 @@ class Tree
       @selection.classed('hx-opened', false)
       @selection.selectAll('.hx-tree-node-children').style('display', 'none')
 
-    animate = if animate? then animate
-    else @options.animate
-
+    animate = if animate? then animate else @options.animate
```
c-frater: Not sure why these were on multiple lines - It didn't really help with readability so I undid it

modules/tree/main/index.coffee
```diff
@@ -10,33 +40,44 @@ formatIcon = (node, iconElement, animate) =>
...
 
-  showDisabled = hx.select(element.parentNode).selectAll('.hx-tree-node').select('.hx-tree-node-children').selectAll('.hx-tree-node').size() > 0
-  showDisabled = if tree.options.hideDisabledButtons then false else showDisabled
+  siblings = hx.select(element.parentNode).shallowSelectAll('.hx-tree-node')
+  siblingsHaveChildren = siblings
+    .shallowSelect('.hx-tree-node-children')
+    .shallowSelectAll('.hx-tree-node')
+    .size() > 0
+
+  siblingsHaveLazyChildren = siblingsHaveChildren or siblings.nodes.map((node) -> hx.component(node)?).some(hx.identity)
```
c-frater: Detect regular children before lazy children - helps speed things up if there are children that have already been renderered

modules/tree/main/index.coffee
```diff
@@ -1,5 +1,35 @@
+createNodeView = (node, renderer, lazy) ->
+  node.view('.hx-tree-node')
+    .enter (d) -> @append(createTreeNode(d, renderer, lazy)).node()
+
+createChildren = (children, renderer, lazy) ->
+  children.map (child) -> createTreeNode(child, renderer, lazy)
+
+createTreeNode = (data, renderer, lazy) ->
```
c-frater: This function now handles the creation of the tree nodes when in JSON mode - it doesn't need to care about the tree because the logic for open/close is in `format` (to retain handling for `html` trees)

modules/tree/main/index.coffee
```diff
@@ -1,5 +1,35 @@
...
+  treeNode = hx.detached('div').class('hx-tree-node')
+
+  nodeContent = hx.detached('div').attr('class', 'hx-tree-node-content')
+  renderer(nodeContent.node(), data)
+  if data.children? and data.children.length > 0
+    childContainer = hx.detached('div').class('hx-tree-node-children').style('display', 'none')
+    if lazy
+      childView = createNodeView(childContainer, renderer, lazy)
+      hx.component.register(treeNode.node(), {
+        renderChildren: -> childView.apply(data.children)
```
c-frater: I couldn't think of a more reliable way to not only detect whether a node has children to render and actually _do_ the render. This is used in `format` -> `renderLazyChildren`

modules/tree/main/index.coffee
```diff
@@ -10,33 +40,44 @@ formatIcon = (node, iconElement, animate) =>
         .go(true)
 
 format = (tree, element, animate) ->
+  treeNode = hx.select(element)
+  treeNodeComponent = treeNode.component()
+
+  renderLazyChildren = (rootNode, recursive) ->
+    if rootNode.component()? and rootNode.selectAll('.hx-tree-node').empty()
+      rootNode.component().renderChildren()
+      formatChildren(tree, rootNode.selectAll('.hx-tree-node'), animate)
```
c-frater: Detect if a node has children to render (i.e. it has `renderChildren` but doesn't have any `hx-tree-node`s) and renders them.
`formatChildren` is a helper so we don't have to refresh the whole tree, only the bit that's changed

modules/tree/main/index.coffee
```diff
@@ -10,33 +40,44 @@ formatIcon = (node, iconElement, animate) =>
...
-
-    if root.style('display') is 'block'
-      selection.forEach (node) ->
-        if node isnt rootNode
-          if node.style('display') isnt 'block'
-            parentNode = node.node().parentNode
-            iconElement = hx.select(parentNode).select('.hx-tree-node-parent-icon').node()
-            node.style('display', 'block')
-            formatIcon(parentNode, iconElement, tree.options.animate)
+    renderLazyChildren(treeNode, true)
```
c-frater: We recursively load the children here - this is intentionally 'blocking' so we can use the same logic to do the double click opening as previous

modules/tree/main/index.coffee
```diff
@@ -10,33 +40,44 @@ formatIcon = (node, iconElement, animate) =>
...
+    treeNode.selectAll('.hx-tree-node-children').style('display', 'block')
+
+    if treeNode.style('display') is 'block'
+      formatChildren(tree, treeNode.selectAll('.hx-tree-node'), animate)
```
c-frater: This is probably what was causing the rendering issue I was having when opening a lot of nodes - The old functionality worked better...

modules/tree/main/index.coffee
```diff
@@ -82,22 +123,20 @@ class Tree
     @options = hx.merge.defined {
       hideDisabledButtons: false
       animate: true
-      renderer: (elem, data) -> hx.select(elem).html(data)
+      renderer: (elem, data) -> hx.select(elem).html(data.name || data)
       items: []
     }, options
 
-    @selection = hx.select(@selector).classed('hx-openable', true)
+    @selection = hx.select(@selector).classed('hx-tree hx-openable', true)
```
c-frater: Not sure why we weren't adding `hx-tree` here - figured it makes sense to add it

modules/tree/main/index.coffee
c-frater: Thinking about it, the reason this is slow is because it's using a `view` to render the children - Could we instead change `view` to only render everything before appending? (if it doesn't already do that)

PR #315, opened 2016-10-03
#254 update the default timezone for hx.preferences when not using moment

### LINKED ISSUE #254: Preferences: should attempt to guess the correct timezone
Currently, if the application isn't using `Moment` for timezone lookup, the timezone is always set to `UTC+00:00`. It should attempt to work out the timezone of the user by default and apply that.

An example method for looking up the timezone (and converting to the correct format) would be: 

```js
function getTimezone (offset) {
  var zeroPad = hx.format.zeroPad(2)
  var modifier = offset > 1 ? '-' : '+'
  var absOffset = Math.abs(offset)
  var minutes = Math.abs(absOffset % 60)
  var hours = (absOffset - minutes) / 60
  return 'UTC' + modifier + zeroPad(hours) + ':' + zeroPad(minutes)
}

getTimezone((new Date()).getTimezoneOffset()) // returns 'UTC+01:00' for BST
```

c-frater:
resolves #254

### DIFF
--- a/modules/preferences/main/index.coffee
+++ b/modules/preferences/main/index.coffee
@@ -143,6 +143,14 @@ LocalStoragePreferencesStore = {
 
 lookupLocale = (locale) -> localeList.filter((l) -> l.value.toLowerCase() is locale.toLowerCase())[0]
 
+zeroPad = hx.format.zeroPad(2)
+defaultTimezoneLookup = (offset) ->
+  modifier = if offset > 1 then '-' else '+'
+  absOffset = Math.abs(offset)
+  minutes = absOffset % 60
+  hours = (absOffset - minutes) / 60
+  "UTC#{modifier}#{zeroPad(hours)}:#{zeroPad(minutes)}"
+
 class Preferences extends hx.EventEmitter
   constructor: ->
     super
@@ -230,13 +238,7 @@ class Preferences extends hx.EventEmitter
         -(moment.tz.zone(timezone).offset(timestamp) / 60)
       @timezone guessedMomentTimezone
     else
-      zeroPad = hx.format.zeroPad(2)
-      offset = (new Date()).getTimezoneOffset()
-      modifier = if offset > 1 then '-' else '+'
-      absOffset = Math.abs(offset)
-      minutes = absOffset % 60
-      hours = (absOffset - minutes) / 60
-      @timezone "UTC#{modifier}#{zeroPad(hours)}:#{zeroPad(minutes)}"
+      @timezone defaultTimezoneLookup((new Date()).getTimezoneOffset())
 
   timezone: (timezone) ->
     if arguments.length > 0
@@ -318,3 +320,7 @@ class Preferences extends hx.EventEmitter
 
 hx.preferences = new Preferences
 hx.preferences.localStorageStore = LocalStoragePreferencesStore
+
+hx._.preferences = {
+  defaultTimezoneLookup: defaultTimezoneLookup
+}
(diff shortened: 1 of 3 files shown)


modules/preferences/main/index.coffee
```diff
@@ -230,7 +230,13 @@ class Preferences extends hx.EventEmitter
...
       @timezone guessedMomentTimezone
     else
-      @timezone 'UTC+00:00'
+      zeroPad = hx.format.zeroPad(2)
+      offset = (new Date()).getTimezoneOffset()
+      modifier = if offset > 1 then '-' else '+'
+      absOffset = Math.abs(offset)
+      minutes = absOffset % 60
+      hours = (absOffset - minutes) / 60
+      @timezone "UTC#{modifier}#{zeroPad(hours)}:#{zeroPad(minutes)}"
```
c-frater: I can't think of a better way to calculate the timezone and convert it to `UTC(+/-)HH:MM`

modules/preferences/main/index.coffee
```diff
@@ -230,7 +230,13 @@ class Preferences extends hx.EventEmitter
         -(moment.tz.zone(timezone).offset(timestamp) / 60)
       @timezone guessedMomentTimezone
     else
-      @timezone 'UTC+00:00'
+      zeroPad = hx.format.zeroPad(2)
+      offset = (new Date()).getTimezoneOffset()
+      modifier = if offset > 1 then '-' else '+'
```
jmsmyth: Why is this compared against 1 rather than 0?

modules/preferences/main/index.coffee
jmsmyth: This feels like the construction of the string should be wrapped up in a separate function, and should have some tests to test that function

modules/preferences/main/index.coffee
c-frater: This is because the timezone is `UTC+00:00` - we want it to be `+` for `0` :)

modules/preferences/main/index.coffee
```diff
@@ -143,6 +143,14 @@ LocalStoragePreferencesStore = {
 
 lookupLocale = (locale) -> localeList.filter((l) -> l.value.toLowerCase() is locale.toLowerCase())[0]
 
+zeroPad = hx.format.zeroPad(2)
+defaultTimezoneLookup = (offset) ->
+  modifier = if offset > 1 then '-' else '+'
```
jmsmyth: If `offset` is 1 then you'd get 'UTC+00:01', but if `offset` is 2 you would have 'UTC-00:02'. I'm not sure which of the two is correct, but I'd hazard a guess that they should have the same sign :-p

modules/preferences/test/spec.coffee
```diff
@@ -75,3 +75,9 @@ describe 'hx-preferences', ->
         spy = chai.spy.on(hx, 'consoleWarning')
         hx.preferences.timezone('America').should.equal(hx.preferences)
         spy.should.have.been.called()
+
+    describe 'defaultTimezoneLookup', ->
+      it 'should get the correct string timezone', ->
+        hx._.preferences.defaultTimezoneLookup(0).should.equal('UTC+00:00')
+        hx._.preferences.defaultTimezoneLookup(10).should.equal('UTC-00:10')
```
jmsmyth: May want to add a test for offsets of `-1`, `1` and `2` as well ;-)

modules/preferences/main/index.coffee
c-frater: Done

modules/preferences/test/spec.coffee
c-frater: Done

PR #318, opened 2016-10-03
Add class `hx-crumbtrail` to Crumbtrail root node.

### LINKED ISSUE #310: Crumbtrail: Setting hx-crumbtrail class on node
Should the crumbtrail node already have this class before the constructor is called ? The docs seem to suggest so. If so, hx.crumbtrail should set it. However if so we might want to push it down into the Crumbtrail constructor as it's a bit non-standard to do it this way around.

alpacaaa:
Simple fix for https://github.com/ocadotechnology/hexagonjs/issues/310! 🍣 
Resolves #310

### DIFF
--- a/modules/crumbtrail/main/index.coffee
+++ b/modules/crumbtrail/main/index.coffee
@@ -4,6 +4,9 @@ class Crumbtrail
     hx.component.register(@selector, this)
     self = this
 
+    section = hx.select(@selector)
+    section.classed('hx-crumbtrail', true)
+
     @options = hx.merge.defined {
       renderer: (node, data) ->
         hx.select(node).text(data)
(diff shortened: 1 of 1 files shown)


modules/crumbtrail/main/index.coffee
```diff
@@ -4,6 +4,8 @@ class Crumbtrail
     hx.component.register(@selector, this)
     self = this
 
+    hx.select(@selector).classed('hx-crumbtrail', true)
```
c-frater: If you stored this as `selection` you could re-use it below where we define the `@view` (https://github.com/alpacaaa/hexagonjs/blob/5d44760a18e4184b208b7e7397da1e01ded7e1c9/modules/crumbtrail/main/index.coffee#L24)

alpacaaa: Missed that one, I updated the commit ;)

PR #322, opened 2016-10-04
#299: add a renderer option to notifications

### LINKED ISSUE #299: Allow HTML content inside notification
Allow custom HTML content inside notification OR allow more properties than just text.

For example with notification's header, main text, and footer we could write:
`hx.notify.warning({header:"Abc", content:"Abc", footer:"Abc"});`
to get this result
image

c-frater:
resolves #299

Example JS:
```js
const message = {
  header: 'Failed to do something',
  body: 'Something failed',
  url: 'somewhere/somewhere-else'
}

const options = {
  renderer: (node, message) => {
    const rows = [
      hx.detached('b').text(message.header),
      hx.detached('span').text(message.body),
      hx.detached('i').class('small').text(message.url)
    ].map((row) => hx.detached('div').add(row))

    hx.select(node).classed('customNotice', true).add(rows)
  }
}

hx.notify.warning(message, options)
hx.notify.warning('Regular Notice')
hx.notify('Regular Notice')
```

CSS:
```
.customNotice {
  font-size: 0.9em;
}

.small {
  opacity: 0.8;
  font-size: 0.9em;
}
```

Screenshot:
screen shot 2016-10-04 at 11 03 55

### DIFF
--- a/modules/notify/main/index.coffee
+++ b/modules/notify/main/index.coffee
@@ -7,6 +7,7 @@ setupNotification = (notification, selection) ->
   content = hx.detached('div').class('hx-notification-content')
   msg = notification.message
 
+  # We check `isNumber` here as in some cases it might be sensible to pass a number as the message (e.g. hx.notify.error(404))
   msgIsString = hx.isString(msg) or hx.isNumber(msg)
 
   msgIsNode = not msgIsString and ((msg instanceof hx.Selection) or (msg instanceof HTMLElement))
--- a/modules/notify/main/index.coffee
+++ b/modules/notify/main/index.coffee
@@ -5,7 +5,16 @@ setupNotification = (notification, selection) ->
       .add(hx.detached('i').class('hx-notification-icon ' + notification.options.icon))
 
   content = hx.detached('div').class('hx-notification-content')
-  notification.options.renderer(content.node(), notification.message)
+  msg = notification.message
+
+  msgIsNode = (msg instanceof hx.Selection) or (msg instanceof HTMLElement)
+  msgIsArrayOfNodes = msgIsNode or hx.isArray(msg) and msg.every (item) ->
+    (item instanceof hx.Selection) or (item instanceof HTMLElement)
+
+  if msgIsNode or msgIsArrayOfNodes
+    content.add(msg)
+  else
+    notification.options.renderer(content.node(), msg)
 
   if notification.options.pinnable
     pin = hx.detached('div')
@@ -85,6 +94,7 @@ togglePin = (notification) -> if notification.pinned then notification.unpin() e
 updatePinnedStatus = (notification) ->
   notification.domPin.classed('hx-notification-pin-pinned', notification.pinned)
 
+defaultRenderer = (node, message) -> hx.select(node).text(message)
 
 class Notification
   constructor: (@manager, @message, options) ->
@@ -93,8 +103,11 @@ class Notification
       cssclass: undefined
       timeout: @manager._.defaultTimeout
       pinnable: true
-      renderer: (node, message) -> hx.select(node).text(message)
+      renderer: defaultRenderer
     }, options
+
+    # Prevent undefined renderers being passed in
+    @options.renderer ?= defaultRenderer
 
     @id = nextId(@manager)
 
--- a/modules/notify/main/index.coffee
+++ b/modules/notify/main/index.coffee
@@ -7,14 +7,38 @@ setupNotification = (notification, selection) ->
   content = hx.detached('div').class('hx-notification-content')
   msg = notification.message
 
-  msgIsNode = (msg instanceof hx.Selection) or (msg instanceof HTMLElement)
+  msgIsString = hx.isString(msg) or hx.isNumber(msg)
+
+  msgIsNode = not msgIsString and ((msg instanceof hx.Selection) or (msg instanceof HTMLElement))
   msgIsArrayOfNodes = msgIsNode or hx.isArray(msg) and msg.every (item) ->
     (item instanceof hx.Selection) or (item instanceof HTMLElement)
 
+  msgIsObject = not msgIsString and not msgIsNode and not msgIsArrayOfNodes and hx.isObject(msg)
+
+  if msgIsString
+    content.text(msg)
   if msgIsNode or msgIsArrayOfNodes
     content.add(msg)
-  else
+  else if msgIsObject and notification.options.renderer
     notification.options.renderer(content.node(), msg)
+  else
+    if msgIsObject
+      hx.consoleWarning(
+        'Notification created using an object with invalid arguments\n',
+        'An object was passed to the notification without a renderer being defined\n'
+        "message:",
+        msg,
+        "\nrenderer:",
+        notification.options.renderer
+      )
+    else
+      hx.consoleWarning(
+        'Notification created using an object with invalid arguments\n',
+        'The notification expected a String, Selection, HTMLElement or an Object with matching renderer but was passed:\n',
+        "message:",
+        msg
+      )
+    content.text('ERROR CONSTRUCTING NOTIFICATION')
 
   if notification.options.pinnable
     pin = hx.detached('div')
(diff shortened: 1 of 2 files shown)


modules/notify/main/index.scss
```diff
@@ -72,9 +72,8 @@ body.hx-modal-open > {
     cursor: pointer;
   }
 
-  .hx-notification-text {
+  .hx-notification-content {
```
c-frater: I updated this because it is no longer just `text`.

modules/notify/main/index.coffee
```diff
@@ -92,6 +91,7 @@ class Notification
       cssClass: undefined
       timeout: @manager._.defaultTimeout
       pinnable: true
+      renderer: (node, message) -> hx.select(node).text(message)
```
c-frater: The renderer behaves exactly the same as our other renderers

modules/notify/main/index.coffee
```diff
@@ -1,33 +1,32 @@
...
   selection
-    .append('div')
-      .class('hx-notification-icon-container hx-notification-close')
-      .on 'click', 'hx.notify', -> notification.close()
-      .append('i')
-        .class('hx-icon hx-icon-close')
+    .add(icon)
+    .add(content)
+    .add(pin)
+    .add(close)
```
c-frater: I updated how the notification is built because it was a little confusing.
All the elements are now built detached and added to the notification at the end.

modules/notify/main/index.coffee
```diff
@@ -1,33 +1,32 @@
...
-        .on 'click', 'hx.notify', -> togglePin(notification)
-
-    notification.domPin.append('i')
-      .attr('class', 'hx-icon hx-icon-thumb-tack')
+    pin = hx.detached('div')
+      .class('hx-notification-icon-container hx-notification-pin')
+      .on('click', 'hx.notify', -> togglePin(notification))
+      .add(hx.detached('i').attr('class', 'hx-icon hx-icon-thumb-tack'))
 
+    notification.domPin = pin
```
c-frater: I split `pin` and `notification.domPin` because it makes it easier to work out what's going on in this function vs outside

modules/notify/main/index.coffee
```diff
@@ -1,33 +1,32 @@
...
+  icon = if notification.options.icon? and notification.options.icon.length > 0
+    hx.detached('div').class('hx-notification-icon-container')
+      .add(hx.detached('i').class('hx-notification-icon ' + notification.options.icon))
 
-  selection
-    .append('div')
-      .class('hx-notification-text')
-      .text(notification.message)
+  content = hx.detached('div').class('hx-notification-content')
+  notification.options.renderer(content.node(), notification.message)
```
c-frater: We always run the renderer here now. 
This assumes the renderer hasn't been set explicitly to undefined - not sure if we need to add handling for that?

modules/notify/main/index.scss
```diff
@@ -72,9 +72,8 @@ body.hx-modal-open > {
...
     word-wrap: break-word;
-    margin: 0;
```
c-frater: I'm pretty sure this is hanging around from when it was using the `Layout` module for the alignment

modules/notify/main/index.coffee
```diff
@@ -1,33 +1,32 @@
...
-      .append('div')
-        .class('hx-notification-icon-container hx-notification-pin')
-        .on 'click', 'hx.notify', -> togglePin(notification)
-
-    notification.domPin.append('i')
-      .attr('class', 'hx-icon hx-icon-thumb-tack')
+    pin = hx.detached('div')
+      .class('hx-notification-icon-container hx-notification-pin')
+      .on('click', 'hx.notify', -> togglePin(notification))
+      .add(hx.detached('i').attr('class', 'hx-icon hx-icon-thumb-tack'))
```
c-frater: I was going to use the fluid api for this but I'm not sure it's worth doing in this case

modules/notify/main/index.coffee
jmsmyth: Might be an idea to handle the undefined renderer when resolving the options (ie revert to the default render when undefined is explicitly passed in)

modules/notify/main/index.coffee
```diff
@@ -42,8 +41,10 @@ redraw = (manager) ->
   view = container.view('.hx-notification')
   view.enter (d) ->
     selection = @append('div')
+    optionalClass = if d.options.cssclass then " #{d.options.cssclass}" else ''
+
     selection
-      .class('hx-notification ' + d.options.cssclass)
+      .class("hx-notification#{optionalClass}")
```
c-frater: Resolves an issue with trailing spaces as well as notifications being given the `undefined` class

c-frater:
Added support for these:
```
// selections
hx.notify(hx.detached('div').text('Something'))

// 'pure' html element
hx.notify(document.createElement('div'))

const rows = [
  hx.detached('b').text('Header'),
  hx.detached('span').text('Content'),
  hx.detached('i').text('Location')
].map((row) => hx.detached('div').add(row))

// arrays of selections/html elements
hx.notify(rows)
```

modules/notify/main/index.coffee
```diff
@@ -1,33 +1,41 @@
...
-        .class('hx-notification-icon-container hx-notification-pin')
-        .on 'click', 'hx.notify', -> togglePin(notification)
+  msgIsNode = (msg instanceof hx.Selection) or (msg instanceof HTMLElement)
+  msgIsArrayOfNodes = msgIsNode or hx.isArray(msg) and msg.every (item) ->
+    (item instanceof hx.Selection) or (item instanceof HTMLElement)
 
-    notification.domPin.append('i')
-      .attr('class', 'hx-icon hx-icon-thumb-tack')
+  if msgIsNode or msgIsArrayOfNodes
+    content.add(msg)
```
c-frater: Added support for passing selections/dom nodes in directly

modules/notify/main/index.coffee
```diff
@@ -1,33 +1,65 @@
...
+        notification.options.renderer
+      )
+    else
+      hx.consoleWarning(
+        'Notification created using an object with invalid arguments\n',
+        'The notification expected a String, Selection, HTMLElement or an Object with matching renderer but was passed:\n',
+        "message:",
+        msg
+      )
+    content.text('ERROR CONSTRUCTING NOTIFICATION')
```
c-frater: I'm not sure how best to handle the error case with the notification construction.
If an error is logged, we don't populate the notification and it then has no content. I thought doing `selection.display('none')` would be reasonable enough but ended up having this instead as it makes it more obvious that something has gone wrong and that there _should_ have been a notification.

modules/notify/main/index.coffee
```diff
@@ -1,33 +1,65 @@
...
+        msg,
+        "\nrenderer:",
+        notification.options.renderer
+      )
+    else
+      hx.consoleWarning(
+        'Notification created using an object with invalid arguments\n',
+        'The notification expected a String, Selection, HTMLElement or an Object with matching renderer but was passed:\n',
+        "message:",
+        msg
```
c-frater: Decided to duplicate the call to `hx.consoleWarning` to remove the need for logic switching between messages and what to display in the error

modules/notify/main/index.coffee
```diff
@@ -1,33 +1,65 @@
...
-      .class('hx-notification-text')
-      .text(notification.message)
+  icon = if notification.options.icon? and notification.options.icon.length > 0
+    hx.detached('div').class('hx-notification-icon-container')
+      .add(hx.detached('i').class('hx-notification-icon ' + notification.options.icon))
+
+  content = hx.detached('div').class('hx-notification-content')
+  msg = notification.message
+
+  msgIsString = hx.isString(msg) or hx.isNumber(msg)
```
c-frater: I'm not sure if we _need_ to cope with this but this basically allows `hx.notify(1)` (something you can currently do)

jmsmyth: May want to add a comment to explain why the isNumber is there...

PR #326, opened 2016-10-05
#203: update the tag input to deal with when it is inside a form

### LINKED ISSUE #203: Form Builder / Tag Input: validator behaves oddly for tag inputs in forms
```js
const tagInput = hexagon.tagInput({
  validator: (term) => 'bob'
})

const formSel = hexagon.detached('div')

void new hexagon.Form(formSel.node())
  .addTagInput('Tag Input', {
    tagInputOptions: {
      validator: (term) => 'bob'
    }
  })
  .addSubmit('Submit')

const bodySel = hexagon.select('body')
bodySel.append(tagInput)
bodySel.append(formSel)
```

In the first case, pressing enter shows the validation error.

In the second case, pressing enter doesn't show any validation error, it just refuses to create a tag and shows a thick black box around the tag input.

(The error will show if you press the submit button of the whole form, but this is rather counterintuitive).

c-frater:
Also update the styles so the focus/invalid borders show correctly and update the `validateForm` function so that it shows the error for the focused element if applicable.
Resolves #203

Example:
```js
const tagInput = hx.tagInput({
  validator: (term) => 'bob'
})

const formSel = hx.detached('form')

void new hx.Form(formSel.node())
  .addText('Text', {required: true})
  .addTagInput('Tag Input', {
    tagInputOptions: {
      validator: (term) => 'bob'
    }
  })
  .addSubmit('Submit')

const bodySel = hx.select('body')
bodySel.append(tagInput)
bodySel.append(formSel)
```

Notes:
- Entering a tag and pressing enter shows the inbuilt browser error when outside the form and the `validateForm` error when used inside a form
- Validation is now based on which element is focused (e.g. leave 'required' field empty and try to enter a tag)
- Other behaviour is as previous

### DIFF
--- a/modules/tag-input/main/index.coffee
+++ b/modules/tag-input/main/index.coffee
@@ -44,8 +44,17 @@ class TagInput extends hx.EventEmitter
     if @options.draggable
       _.dragContainer = new hx.DragContainer(@tagContainer.node())
 
-    @form = @selection.append('form')
-    @input = @form.append('input').attr('placeholder', @options.placeholder)
+    isInsideForm = not @selection.closest('form').empty()
+
+    inputContainer = @selection.append(if isInsideForm then 'div' else 'form')
+      .class('hx-tag-input-container')
+
+    validationForm = if isInsideForm
+      @selection.closest('.hx-form')
+    else
+      inputContainer
+
+    @input = inputContainer.append('input').attr('placeholder', @options.placeholder)
     if @options.autocompleteData?
       isValid = if @options.validator? then (item) => not @options.validator(item) else hx.identity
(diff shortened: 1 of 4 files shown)


modules/tag-input/main/index.coffee
```diff
@@ -44,8 +44,17 @@ class TagInput extends hx.EventEmitter
     if @options.draggable
       _.dragContainer = new hx.DragContainer(@tagContainer.node())
 
-    @form = @selection.append('form')
-    @input = @form.append('input').attr('placeholder', @options.placeholder)
+    isInsideForm = not @selection.closest('form').empty()
+
+    inputContainer = @selection.append(if isInsideForm then 'div' else 'form')
```
jmsmyth: nice!

PR #331, opened 2016-10-06
#328: resolve the issue with the number picker

### LINKED ISSUE #159: Number Picker: disable the decrement/increment buttons when the value is at the min/max
This involves adding/removing the disabled attribute to the buttons when the value changes - currently the buttons just ignore input when they hit the limits

### LINKED ISSUE #328: NumberPicker: value is undefined in change event when typing in values
In Hexagon `1.5.1`:
- type any number - you receive single `change` event with that value when you leave the input
- type incorrect value - it is fixed when you leave the input

In Hexagon `1.7.0` (and probably `1.6.0`):
- type any number - you receive single `change` event with `undefined` as a value when you leave the input
- type incorrect value - it stays when you leave the input

c-frater:
Also, disable the buttons when the number picker reaches the min/max value 
Resolves #328 
Edit: also resolves #159 - I didn't realise there was an issue for it and just made the addition...

### DIFF
--- a/modules/number-picker/main/index.coffee
+++ b/modules/number-picker/main/index.coffee
@@ -1,10 +1,9 @@
-checkValue = (numberPicker, context) ->
-  value = oldValue = context.value()
-  max = numberPicker.max()
-  min = numberPicker.min()
+checkValue = (value, min, max) ->
   if max isnt undefined then value = Math.min(value, max)
   if min isnt undefined then value = Math.max(value, min)
-  if value isnt oldValue then context.value(value)
+  value
+
+getDisabled = (disabled, val, edge) -> if disabled or (val is edge) then 'disabled' else undefined
 
 addHoldHandler = (incrementOnHold, incrementDelay, selection, incrementFn) ->
   if incrementOnHold
@@ -62,8 +61,9 @@ class NumberPicker extends hx.EventEmitter
     @selectInput = selection.append('input')
     @selectInput.attr('type', 'number')
     @selectInput.on 'blur', 'hx.number-picker', =>
-      @emit 'input-change', {value: @value()}
-      @value(undefined, @selectInput.value())
+      if @selectInput.attr('readonly') is undefined
+        @value(undefined, @selectInput.value())
+        @emit 'input-change', {value: @value()}
 
     decrementButton = selection.append('button').attr('type', 'button').class('hx-number-picker-decrement hx-btn ' + @options.buttonClass)
     decrementButton.append('i').class('hx-icon hx-icon-chevron-down')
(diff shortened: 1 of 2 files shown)


modules/number-picker/main/index.coffee
```diff
@@ -72,27 +72,33 @@ class NumberPicker extends hx.EventEmitter
...
+        # If both value and screen value are defined, we set the input text to be the screen value
+        # otherwise we use the provided value (as it is set via the input)
+        .value(if value? and screenValue? then screenValue else newVal)
+
+      selection = hx.select(@selector)
+      selection.select('.hx-number-picker-decrement')
+        .attr('disabled', getDisabled(@options.disabled, newVal, @min()))
+
+      selection.select('.hx-number-picker-increment')
+        .attr('disabled', getDisabled(@options.disabled, newVal, @max()))
```
c-frater: Set the disabled state of the buttons whenever the value is updated

modules/number-picker/main/index.coffee
```diff
@@ -72,27 +72,33 @@ class NumberPicker extends hx.EventEmitter
...
   value: (value, screenValue) ->
     if arguments.length > 0
       prevValue = @value()
-      newVal = if not value? and screenValue then screenValue else value
 
-      if @_.max isnt undefined and newVal > @_.max then newVal = @_.max
-      if @_.min isnt undefined and newVal < @_.min then newVal = @_.min
+      valueToUse = if not value? and screenValue then Number(screenValue) else value
+
+      newVal = checkValue(valueToUse, @min(), @max())
```
c-frater: I could probably use hx.clamp for this?

modules/number-picker/main/index.coffee
```diff
@@ -72,27 +72,33 @@ class NumberPicker extends hx.EventEmitter
     if @options.max isnt undefined then @max @options.max
     if @options.min isnt undefined then @min @options.min
     if @options.disabled then @disabled(@options.disabled)
-
-    @selectInput
-      .attr('data-value', @options.value)
-      .value(@options.value)
+    @value @options.value
```
c-frater: We were setting the value from the constructor using the input which bypasses the checks inside the value method.
We may want to suppress the event here although you can't listen to it before this is fired so I'd say this is probably fine.

modules/number-picker/main/index.coffee
```diff
@@ -62,8 +61,9 @@ class NumberPicker extends hx.EventEmitter
     @selectInput = selection.append('input')
     @selectInput.attr('type', 'number')
     @selectInput.on 'blur', 'hx.number-picker', =>
-      @emit 'input-change', {value: @value()}
-      @value(undefined, @selectInput.value())
+      if @selectInput.attr('readonly') is undefined
```
c-frater: The events were being fired when the input was readonly, causing issues with localised number values (e.g. time picker)

modules/number-picker/main/index.coffee
jmsmyth: It would seem that hx.clamp doesn't handle undefined for min or max at the moment, so hx.clamp would need an upgrade to be used here

modules/number-picker/main/index.coffee
```diff
@@ -101,7 +107,7 @@ class NumberPicker extends hx.EventEmitter
     if arguments.length > 0
       @_.min = val
       @selectInput.attr('min', val)
-      checkValue(this, this)
+      @value(@value())
```
jmsmyth: This looks odd.. Is is just to get events to fire or something?

c-frater: It's used to `clamp` the value - I guess we could check here if the value is smaller than the min and then do this but I figured as we do the check inside `@value` anyway it made sense to use it.

PR #337, opened 2016-10-11
#236: properly hide sticky table cells so they don't show in ctrl+f results

### LINKED ISSUE #236: Data Table: Sticky table shows up in Ctrl-F search results
Apparently, this should be solved by setting the visibility of the sticky table's body to hidden

c-frater:
## Description

Updated the internals to be more deliberate when hiding/showing cells. This involved making all the cells (`td, th`) in cloned tables invisible (using `hx-sticky-table-invisible`) by default and then only making them visible when we create the sticky headers.

This update means that all cells that cannot be seen when scrolling are given the `visibility: hidden` style which prevents them showing in `ctrl+f` results.

I did this in a common way and removed some duplicate code by pulling it out into a function.

I also updated a few places  internally to use `shallowSelect`/`shallowSelectAll` as it is more efficient and removed the need for some code.

## Motivation and Context

Resolves #236 

## How Has This Been Tested?

Tested in-browser using the code from https://www.hexagonjs.io/examples/sticky-table-headers/ and making sure that there are only `567` search results for `Cell Text` and `61` search results for `Head Text` to match the HTML.
Also made sure `dataTable` tests still pass as they use the sticky table headers.

N/A

## Types of changes

## Checklist:

- [ ] I have added tests to cover my changes.

### DIFF
--- a/modules/sticky-table-headers/main/index.coffee
+++ b/modules/sticky-table-headers/main/index.coffee
@@ -12,20 +12,19 @@ updateScrollIndicators = (wrapper, top, right, bottom, left) ->
   bottom.style('display', if canScrollDown then 'block' else '')
   left.style('display', if canScrollLeft then 'block' else '')
 
-updateHeaderPositions = (container) ->
+updateHeaderPositions = (container, wrapperNode) ->
   # Set the relative positions of the two headers.
   # This method reduces flickering when scrolling on mobile devices.
-  node = getChildren(container, '.hx-sticky-table-wrapper')[0]
-  leftOffset = - node.scrollLeft
-  topOffset = - node.scrollTop
+  leftOffset = - wrapperNode.scrollLeft
+  topOffset = - wrapperNode.scrollTop
 
-  topNode = getChildren(container, '.hx-sticky-table-header-top')[0]
+  topNode = container.shallowSelect('.hx-sticky-table-header-top')
   if topNode?
-    hx.select(topNode).select('.hx-table').style('left', leftOffset + 'px')
+    topNode.select('.hx-table').style('left', leftOffset + 'px')
 
-  leftNode = getChildren(container, '.hx-sticky-table-header-left')[0]
+  leftNode = container.shallowSelect('.hx-sticky-table-header-left')
   if leftNode?
-    hx.select(leftNode).select('.hx-table').style('top', topOffset + 'px')
+    leftNode.select('.hx-table').style('top', topOffset + 'px')
 
 
 cloneEvents = (elem, clone) ->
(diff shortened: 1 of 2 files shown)


modules/sticky-table-headers/main/index.coffee
```diff
@@ -63,7 +62,18 @@ getChildren = (parent, selector, single) ->
...
   for i in [0...real.length]
     cloneEvents(real[i], cloned[i])
+    # The real table shouldn't show the sticky header nodes
     hx.select(real[i]).classed('hx-sticky-table-invisible', true)
+    hx.select(cloned[i]).classed('hx-sticky-table-invisible', false)
+
+cloneTableAndNodeEvents = (selection, realTable, tableClone, body, single) ->
+  innerTableClone = selection.append tableClone.clone(true)
+  # We make all the cells invisible by default and then only make the ones we care about visible in `createStickyHeaderNodes`
+  innerTableClone.selectAll('th, td').classed('hx-sticky-table-invisible', true)
```
c-frater: The key change is here, where we make _all_ the cells invisible in the 'cloned' table and then in `createStickyHeaderNodes` make only the ones we care about visible again.

modules/sticky-table-headers/main/index.coffee
```diff
@@ -63,7 +62,18 @@ getChildren = (parent, selector, single) ->
...
+  realNodes = getChildrenFromTable realTable, body, single
+  clonedNodes = getChildrenFromTable innerTableClone, body, single
```
c-frater: I attempted to replace `getChildrenFromTable` with `shallowSelect`/`shallowSelectAll` but it caused issues with events not being cloned properly (possibly only in `PhantomJS`):

```coffee
getChildrenFromCompoundSelector = (table, selectorChain) ->
  selectorChain.reduce (sel, [selector, single]) ->
    sel[if single then 'shallowSelect' else 'shallowSelectAll'](selector)
  , table
```
Use:
```coffee
getChildrenFromCompoundSelector(table, [['thead', true], ['tr'], ['th, td']])
```

There must be a better way of doing `getChildrenFromTable`...

modules/sticky-table-headers/main/index.coffee
```diff
@@ -12,20 +12,19 @@ updateScrollIndicators = (wrapper, top, right, bottom, left) ->
   bottom.style('display', if canScrollDown then 'block' else '')
   left.style('display', if canScrollLeft then 'block' else '')
 
-updateHeaderPositions = (container) ->
+updateHeaderPositions = (container, wrapperNode) ->
   # Set the relative positions of the two headers.
   # This method reduces flickering when scrolling on mobile devices.
-  node = getChildren(container, '.hx-sticky-table-wrapper')[0]
```
c-frater: Replacing this with `shallowSelect` made some of the data table tests fail (as it couldn't find `hx-sticky-table-wrapper) so I passed it into the function instead - I think it was because `PhantomJS` doesn't 100% correctly with detached nodes (as it worked fine in the browser)

modules/sticky-table-headers/main/index.coffee
```diff
@@ -63,7 +62,18 @@ getChildren = (parent, selector, single) ->
 createStickyHeaderNodes = (real, cloned) ->
   for i in [0...real.length]
     cloneEvents(real[i], cloned[i])
+    # The real table shouldn't show the sticky header nodes
     hx.select(real[i]).classed('hx-sticky-table-invisible', true)
+    hx.select(cloned[i]).classed('hx-sticky-table-invisible', false)
+
+cloneTableAndNodeEvents = (selection, realTable, tableClone, body, single) ->
+  innerTableClone = selection.append tableClone.clone(true)
```
c-frater: Cloning the nodes/events in a detached state doesn't seem to work so we have to append the clone here.

PR #339, opened 2016-10-11
#313: update tag input to use more of the autocomplete internal

### LINKED ISSUE #313: Tag Input: invalid tag entered when using autocomplete
When using a tag input with autocomplete enabled, it is possible to enter tags incorrectly by _clicking_ the value in the autocomplete, as opposed to using the keyboard.

Example:
```js
var autocompleteData = [
  'bob',
  'steve',
  'dave'
]

hx.select('body').clear()
  .add(hx.tagInput({
    autocompleteData: autocompleteData
  }))
```
- type `b`
- click `bob` entry (must click, entering using the keyboard works fine)
- `b` tag is entered, along with `bob` tag

I think this is because the `blur` event causes the text entered to be added as a tag and then the autocomplete selected value is also added as a tag afterwards. This should be resolvable by either disabling the `blur` event when autocomplete is enabled or by adding handling to detect when an autocomplete item has been selected and not add the manually entered text.

c-frater:
## Description

Hand over the handling for blur/enter to autocomplete as it deals nicely with the `mustMatch` functionality by removing the `blur` event and only calling the `add` method inside the tag input when the autocomplete does not exist or has been updated.

## Motivation and Context

Resolves #313 

## How Has This Been Tested?

Added a regression test and tested in browser using example in #313 

N/A

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/modules/tag-input/main/index.coffee
+++ b/modules/tag-input/main/index.coffee
@@ -72,10 +72,11 @@ class TagInput extends hx.EventEmitter
       if event.keyCode is 13
         if @form.node().checkValidity()
           event.preventDefault()
-          name = @input.value()
-          if name
-            _.userEvent = true
-            @add name
+          if not @_.autocomplete
+            name = @input.value()
+            if name
+              _.userEvent = true
+              @add name
 
     @input.on 'input', 'hx.tag-input', hasError
 
@@ -87,9 +88,11 @@ class TagInput extends hx.EventEmitter
         if @input.value() is ''
           selection = @tagContainer.selectAll('.hx-tag')
           if selection.size() > 0
+            @_.autocomplete?.hide()
             nodeSelection = hx.select(selection.node(selection.size()-1))
             value = nodeSelection.text()
             nodeSelection.remove()
+            @_.autocomplete?.show()
             @emit 'remove', {value: value, type: 'user'}
 
     @input.on 'keyup', 'hx.tag-input', (event) ->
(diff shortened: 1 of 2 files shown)


modules/tag-input/main/index.coffee
```diff
@@ -87,19 +88,22 @@ class TagInput extends hx.EventEmitter
...
     @input.on 'keyup', 'hx.tag-input', (event) ->
       if (event.keyCode or event.charCode) is 8
         backspacedown = false
         true
 
-    @input.on 'blur', 'hx.tag-input', (event) =>
-      if @input.value().length > 0 and not hasError()
-        @add(@input.value(), undefined)
+    if not @_.autocomplete
+      @input.on 'blur', 'hx.tag-input', (event) =>
```
c-frater: We don't need the blur event at all when we have an autocomplete as `AutoComplete` deals with it

modules/tag-input/test/spec.coffee
```diff
@@ -236,3 +236,23 @@ describe 'tag-input', ->
...
+      ti.input.value('b')
+      testHelpers.fakeNodeEvent(ti.input.node(), 'input')({})
+      clock.tick(dropdownAnimateDuration)
+
+      dropdown = ti._.autocomplete._.menu.dropdown._.dropdown
+      target = dropdown.select('.hx-menu-item').node()
+
+      testHelpers.fakeNodeEvent(ti.input.node(), 'blur')({})
+      testHelpers.fakeNodeEvent(dropdown.node(), 'click')({target: target})
+      ti.items().should.eql(['bob'])
```
c-frater: Tested this returns `['b','bob']` before the changes and realised that the dropdown event is fired in a weird way which made it awkward to test as you have to pass `target` through to the event. Might be useful if you need to test other menu-related things.

georgesimms:
Is it appropriate for modules to use internal parts of modules that end users are not intended to use ? In particular, will this work with webpack well ?

c-frater:
In this case, we're just relying on the autocomplete doing the right thing as opposed to explicitly requiring any functionality/shared code which will work fine with webpack.

I generally try to avoid using internal stuff (e.g. anything on `_` from outside the current module) but sometimes you have to hook into other modules to override/detect things that are otherwise inaccessible.

PR #340, opened 2016-10-12
#179: bugfix event emitter memory leak

### LINKED ISSUE #179: EventEmitter: memory leak when many unique namespaces are used
When removing a handler from an event emitter the underlying BasicEventEmitter reference is held onto, even when it contains no handlers. When removing handlers, check what can be cleaned up to avoid memory leaks.

This is a problem when many generated unique namespaces are used, as over time it gathers up more and more BasicEventEmitters.

c-frater:
## Description

Updated the `off` method of `EventEmitter` to check if a `BasicEventEmitter` is empty after the callback/listeners have been removed and remove it from the `EventEmitter` if it has no listeners left.

## Motivation and Context

Resolves #179 

## How Has This Been Tested?

In addition to adding tests, I used the following to test in the browser:
```js
hx.range(500).map(() => setInterval(addTest, 1))
function addTest () {
  const cd = new hx.ClickDetector()
  cd.cleanUp()
}
```
This rapidly creates name-spaced event emitters on the `document` (for `pointer-up`/`pointer-down`) and then calls the cleanup on the click detector to remove them straight away.
I then used the chrome developer tools to look a the JS Heap over 6 seconds (that's all the buffer can store in this example):

Before Change (bottom of graph is 240mb, top is 366mb):
screen shot 2016-10-12 at 11 36 38
After Change (bottom of graph is 11mb, top is 33mb)
screen shot 2016-10-12 at 11 35 41

Before the change, the memory use increased rapidly at a constant rate until the browser tab crashed.
After the change, the memory usage stays consistently low and does not crash the browser tab.

N/A

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/modules/event-emitter/main/index.coffee
+++ b/modules/event-emitter/main/index.coffee
@@ -9,7 +9,7 @@ class BasicEventEmitter
     cb(name, data) for cb in @allCallbacks.entries()
     this
 
-  isEmpty: -> hx.sum(@callbacks.values().map((list) -> list.size)) is 0 and @allCallbacks.size is 0
+  isEmpty: -> @callbacks.values().every((list) -> list.size is 0) and @allCallbacks.size is 0
 
   # register a callback against the name given
   on: (name, callback) ->
--- a/modules/event-emitter/main/index.coffee
+++ b/modules/event-emitter/main/index.coffee
@@ -1,6 +1,4 @@
-
 class BasicEventEmitter
-
   constructor: ->
     @callbacks = new hx.Map
     @allCallbacks = new hx.List
@@ -10,6 +8,8 @@ class BasicEventEmitter
     if @callbacks.has(name) then cb(data) for cb in @callbacks.get(name).entries()
     cb(name, data) for cb in @allCallbacks.entries()
     this
+
+  isEmpty: -> hx.sum(@callbacks.values().map((list) -> list.size)) is 0 and @allCallbacks.size is 0
 
   # register a callback against the name given
   on: (name, callback) ->
@@ -74,19 +74,23 @@ class EventEmitter
     ee.emitters.add(be)
     be
 
-  removeEmitter = (ee, namespace) ->
-    if ee.emittersMap.has(namespace)
+  removeEmitter = (ee, be, namespace) ->
+    if namespace and ee.emittersMap.has(namespace)
       ee.emittersMap.delete(namespace)
-      ee.emitters.remove(ee)
-    ee
+    else
+      lookedUpNamespace = ee.emittersMap.entries().filter(([_, e]) -> e is be)[0]?[0]
+      if lookedUpNamespace
+        ee.emittersMap.delete(lookedUpNamespace)
+    ee.emitters.remove(be)
 
   # emit an object to all callbacks registered with the name given
   emit: (name, data) ->
     if not @suppressedMap.get(name)
-      if @deprecatedEvents?
-        for e of @deprecatedEvents
-          if @deprecatedEvents[e].event is name
-            @emit e, data
+      # XXX: Deprecated event check - This is useful to have if we need to deprecated events in the future
+      # if @deprecatedEvents?
+      #   for e of @deprecatedEvents
+      #     if @deprecatedEvents[e].event is name
+      #       @emit e, data
 
       for emitter in @emitters.entries()
         emitter.emit(name, data)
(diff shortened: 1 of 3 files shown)


modules/selection/test/spec.coffee
```diff
@@ -1078,15 +1078,15 @@ describe 'Selection Api', ->
     div = hx.detached('div')
     called = false
     div.on 'click', (e) -> called = true
-    div.node().__hx__.eventEmitter.emit('click', {})
+    testHelpers.fakeNodeEvent(div.node(), 'click')({})
```
c-frater: One of the tests was failing for selection and whilst I was investigating I thought I might as well update to use `testHelpers`

modules/event-emitter/main/index.coffee
```diff
@@ -130,10 +134,22 @@ class EventEmitter
...
+      if not callback and not hx.isString(namespace)
+        callback = namespace
+
+      emitters = @emitters.entries()
+      emittersToRemove = []
+      emitters.forEach (emitter) =>
         emitter.off(name, callback)
+        if emitter isnt @global and emitter.isEmpty()
+          emittersToRemove.push(emitter)
+      emittersToRemove.map((e) => removeEmitter(this, e))
```
c-frater: This was originally using a for loop - I changed to `forEach` as it seems more reliable in this case.
I also split the removal out of the function as removing inside causes reference issues and 'leftover' emitters when doing `.off()` and using several namespaces.

modules/event-emitter/main/index.coffee
```diff
@@ -74,19 +74,23 @@ class EventEmitter
...
+  removeEmitter = (ee, be, namespace) ->
+    if namespace and ee.emittersMap.has(namespace)
       ee.emittersMap.delete(namespace)
-      ee.emitters.remove(ee)
-    ee
+    else
+      lookedUpNamespace = ee.emittersMap.entries().filter(([_, e]) -> e is be)[0]?[0]
+      if lookedUpNamespace
+        ee.emittersMap.delete(lookedUpNamespace)
+    ee.emitters.remove(be)
```
c-frater: Updated `removeEmitter` to take the basic emitter you want to remove and updated to look for the emitter in the namespaced list as otherwise `.off()` doesn't behave properly, leaving emitters in the map but removing them from the `emitters`.
This function was previously unused...

modules/event-emitter/main/index.coffee
```diff
@@ -74,19 +74,23 @@ class EventEmitter
...
     if not @suppressedMap.get(name)
-      if @deprecatedEvents?
-        for e of @deprecatedEvents
-          if @deprecatedEvents[e].event is name
-            @emit e, data
+      # XXX: Deprecated event check - This is useful to have if we need to deprecated events in the future
+      # if @deprecatedEvents?
+      #   for e of @deprecatedEvents
+      #     if @deprecatedEvents[e].event is name
+      #       @emit e, data
```
c-frater: The other part of the deprecated event check is commented out so I commented this out too

modules/event-emitter/main/index.coffee
```diff
@@ -130,10 +134,22 @@ class EventEmitter
...
-      @emittersMap.get(namespace)?.off(name, callback)
+      if @emittersMap.has(namespace)
+        be = @emittersMap.get(namespace)
+        be.off(name, callback)
+        if be.isEmpty()
+          removeEmitter(this, be, namespace)
     else
-      for emitter in @emitters.entries()
+      if not callback and not hx.isString(namespace)
+        callback = namespace
```
c-frater: Previously, this wasn't working correctly when using event emitters directly as you _had_ to supply a namespace to clear a specific callback.
Selection was working fine because we always pass all 3 arguments through.

modules/event-emitter/main/index.coffee
```diff
@@ -11,6 +9,8 @@ class BasicEventEmitter
     cb(name, data) for cb in @allCallbacks.entries()
     this
 
+  isEmpty: -> hx.sum(@callbacks.values().map((list) -> list.size)) is 0 and @allCallbacks.size is 0
```
c-frater: Added function so we can easily check if a basic emitter is empty

jmsmyth: You could avoid the extra array creation here with `every`:
```isEmpty: -> @callbacks.values().every((list) -> list.size is 0) and @allCallbacks.size is 0```

Issue #217, opened 2016-04-20
All: events should be standardised


jmsmyth:
Events should always be objects. They should (roughly) take the form { cause, value, ...}

PR #342, opened 2016-10-14
#311: tabs refactoring

### LINKED ISSUE #311: Tabs: should set up the elements with .hx-tab itself
Eg something like

```
new hx.Tabs('#tabs-script', {
  items: [
    {
      title: 'Tabs 1',
      content: 'Tab 1 selected'
    },
    {
      title: 'Tabs 2',
      content: 'Tab 2 selected',
      context: 'action'
    }
  ]
})
```

This can be done in a backwards compatible way by using the div that already exists if any and if not then creating it

georgesimms:
## Description

Refactor tabs module
## Motivation and Context

Resolves #311 
- Adds fluid support to tabs
- Makes tabs API more like everything else in hexagon
- Removes the requirement from the caller to do tedious internal html setup in tabs - allows the exact nature of the html to be refactored more easily in the future
## How Has This Been Tested?

Before making the change I added tests for the existing behavior (previously there were no tests).
New tests added for new behavior.

No UI changes
## Checklist:


modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,116 @@
...
   selection.selectAll('.hx-tab').classed('hx-tab-active', false)
   hx.select(element).classed('hx-tab-active', true)
 
   context = hx.palette.context(element)
-  hx.palette.borderContext(selection.select('.hx-tabs-content'), context)
+  tabsContent = selection.select('.hx-tabs-content')
+  hx.palette.borderContext(tabsContent, context)
 
-  hx.select(tabs.selector).selectAll('.hx-tab-content').classed('hx-tab-content-hidden', true)
-  hx.select('#' + hx.select(element).attr('data-content')).classed('hx-tab-content-hidden', false)
```
georgesimms: Although element IDs should be unique in the document, this won't work if the tabs are in a detached div

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,116 @@
...
+  rootSelection.selectAll('.hx-tab-content')
+    .classed('hx-tab-content-hidden', true)
+  item = tabs.items()[i]
+  if item?
+    tabs._.options.contentRenderer tabsContent.node(), item.content
+  else
+    tabToSelectSelector = '#' + hx.select(element).attr('data-content')
+    tabToSelect = rootSelection.select(tabToSelectSelector)
+    tabToSelect.classed('hx-tab-content-hidden', false)
+  tabs.emit('change', {id: i, value: i, cause})
```
georgesimms: See #217 - this should be backwards compatible

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,116 @@
...
+      items: []
+      titleRenderer: defaultRenderer
+      contentRenderer: defaultRenderer
+    }
+
+    resolvedOptions = hx.merge defaultOpts, options
+
+    @_ = {}
+
+    @_.options = resolvedOptions
```
georgesimms: we need to do this here, so the items setter can access it on line 63

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,116 @@
...
+      this
+
+    else
+      @_.options.items
+
   select: (i, force) ->
     if @selected != i or force
       tab = hx.select(@selector).selectAll('.hx-tab').nodes[i]
-      onTabSelected(this, tab, i)
+      onTabSelected(this, tab, i, 'api')
```
georgesimms: update this to use the "cause" property

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,116 @@
...
+          onTabSelected self, node.node(), i, 'user'
 
     # make the first tab active
     @select(0)
 
+
+  items: (newItems) ->
+    if arguments.length
+      if newItems.length is 0
+        hx.consoleWarning 'Setting items to empty array'
```
georgesimms: this is a nonsense thing to do, and if we allow it weird stuff will happen further down the line. In particular, we cannot select any element

modules/tabs/test/spec.coffee
```diff
@@ -1,2 +1,220 @@
...
+
+    titleRenderer = (node, data) -> hx.select(node).text(titles[data])
+
+    contentRenderer = chai.spy (node, item) ->
+      { name, town } = item
+      hx.select node
+        .add hx.detached('span').text name
+        .add hx.detached('span').text town
+
+    hx.tabs { titleRenderer, contentRenderer, items }
```
georgesimms: example of how to use the fluid api

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,116 @@
...
+    rootSel.classed 'hx-tabs', true
+
+
+    titleRenderer = resolvedOptions.titleRenderer
+
+    tabsContent = rootSel.select '.hx-tabs-content'
+
+    if tabsContent.empty()
+      tabsContent = rootSel.append 'div'
+        .class 'hx-tabs-content'
```
georgesimms: ideally this would always be created in the constructor. Detecting if it exists is mostly for backwards compatibility

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,116 @@
...
+      onTabSelected(this, tab, i, 'api')
+
+hx.Tabs = Tabs
 
-hx.Tabs = Tabs
\ No newline at end of file
+hx.tabs = (opts) ->
+  ret = hx.detached 'div'
+  tabs = new hx.Tabs ret.node(), opts
+  ret
```
georgesimms: Does a helper exist for this ?

modules/tabs/test/spec.coffee
```diff
@@ -1,2 +1,212 @@
...
+    { rootSel } = setupTabs()
+    testContext rootSel
+
+  it 'should use the default renderer if none is specified', ->
+    tabs = hx.tabs items: [{ title: 'Title', content: 'Content' }]
+    opts = tabs.component()._.options
+    opts.contentRenderer.should.be.defined
+    opts.titleRenderer.should.be.defined
+    tabs.select('.hx-tabs-content').text().should.equal 'Content'
+    tabs.select('.hx-tab').text().should.equal 'Title'
```
georgesimms: from line 157 to here tests the new behavior

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,113 @@
...
+        node.on 'click', 'hx.tabs', ->
+          onTabSelected self, node.node(), i, 'user'
 
     # make the first tab active
     @select(0)
 
+  items: (newItems) ->
+    if arguments.length
+      if newItems.length is 0
+        hx.consoleWarning 'Setting items to empty array'
```
georgesimms: This is a nonsense thing to do.

Allowing it will just cause weird stuff to happen later on - in particular, you cannot select element 0 if there are 0 elements

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,119 @@
...
+      @_.options.items = newItems
+
+      this
+
+    else
+      @_.options.items
+
+  value: (newValue) ->
+    if arguments.length
+      @select newValue, true
```
georgesimms: not sure what to set `force` to here

modules/tabs/main/index.coffee
```diff
@@ -1,36 +1,120 @@
...
+          tab = hx.detached 'div'
+            .class 'hx-tab'
+          @_.options.titleRenderer tab.node(), title
+          hx.palette.context tab.node(), context
+          tab.on 'click', 'hx.tabs', =>
+            onTabSelected this, tab.node(), i, 'user'
+          tab
+        tabsContent.insertBefore titleBarsToAdd
+
+        @select 0, true
```
georgesimms: force has to be true here, otherwise if you previously selected the first tab it won't get selected here

modules/tabs/main/index.coffee
c-frater: I'd suggest either adding an argument and behaving the same way as `select` or using `false` as it doesn't need to do anything if you've already selected the tab.

demo/index.js
```diff
@@ -57,3 +56,61 @@ new hx.Picker('#viewMenu', {
...
+  var tabs = hx.tabs({
+    items: tabItems,
+    titleRenderer: tabsTitleRenderer,
+    contentRenderer: tabsContentRenderer
+  })
+
+  hx.select('#tabs-fluid').add(tabs)
+}
+
+setupFluidTabs()
```
c-frater: Example script should be in `index.html` in a `script` block after the section it applies to

modules/tabs/test/spec.coffee
```diff
@@ -1,2 +1,212 @@
...
+    rootSel = hx.detached 'div'
+      .add  headerAndBody.map ({ header }) -> header
+      .add tabsContent
+
+    tabs = new hx.Tabs rootSel.node()
+
+    { headerAndBody, rootSel, tabs }
+
+  verifyVisible = (headerAndBody, i) ->
+    expectedVisible = headerAndBody.map (_, index)  -> index is i
```
c-frater: Double spacing :P

modules/tabs/test/spec.coffee
```diff
@@ -1,2 +1,212 @@
...
+
+    actualVisible = headerAndBody.map ({ header }) ->
+      header.classed 'hx-tab-active'
+
+    actualVisible.should.eql expectedVisible
+
+  testChangeEventFired = (cause, tabToSelect, done) ->
```
c-frater: This name is a bit weird - It's not actually testing anything on it's own, you have to call something else to run the test - could this be renamed to `setupChangeEventTest` or something like that?

modules/tabs/test/spec.coffee
```diff
@@ -1,2 +1,212 @@
...
+    testChange 'api', (tabs) -> tabs.component().select 1
+
+  it 'should allow things to be set by the dom', ->
+    testChange 'user', (tabs) -> pretendClickTab tabs, 1
+
+  it 'should set the pallette correctly', ->
+    tabs = setupTabsUsingStructuredObject()
+    headers = tabs.selectAll '.hx-tab'
+    actualContext = headers.map (header) -> hx.palette.context header
+    expectedContext = [undefined, undefined, 'negative', 'action']
```
c-frater: Could we create a shared variable for this and `ctx` used in the setup function?

demo/index.js
georgesimms: What is the point in this file then

modules/tabs/main/index.coffee
georgesimms: changed to false

modules/tabs/test/spec.coffee
georgesimms: done

modules/tabs/test/spec.coffee
georgesimms: fixed

modules/tabs/test/spec.coffee
georgesimms: renamed to defaultContext

demo/index.js
jmsmyth: This file is just for setting up things on the demo page that are not part of the examples (the dropdown that changes the background colour for example). We have put the scripts next to the corresponding piece of html to make these examples easier to read (you wouldn't structure a normal app like this, but since these are just examples, its probably fine)

PR #355, opened 2016-12-09
Disable ‘native’ pointer events

### LINKED ISSUE #354: Pointer Events: PointerEvents added to browsers are not compatible with the hexagon version
## Context

Versions 55+ of Chrome have added a PointerEvent that is called before getting into the 'augmented' pointer events.

https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent use the same `pointerup` etc. names as hexagon uses.

## Possible Fix

Disable the native `PointerEvent` and add it 'intentionally' in another release

## Your Environment

* Hexagon Version: All
* Browser Name and version: Chrome 55+
* Operating System and version (desktop or mobile): All

c-frater:
resolves #354

## Checklist:

### DIFF
--- a/modules/selection/main/index.coffee
+++ b/modules/selection/main/index.coffee
@@ -381,7 +381,7 @@ class Selection
 
       data.listenerNamesRegistered ?= new hx.Set
 
-      if name.indexOf('pointer') is -1 and not data.listenerNamesRegistered.has(name)
+      if name.indexOf('pointer') isnt 0 and not data.listenerNamesRegistered.has(name)
         handler = (e) -> eventEmitter.emit(name, e)
         data.listenerNamesRegistered.add(name)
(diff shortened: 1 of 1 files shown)


modules/selection/main/index.coffee
```diff
@@ -381,7 +381,7 @@ class Selection
 
       data.listenerNamesRegistered ?= new hx.Set
 
-      if not data.listenerNamesRegistered.has(name)
+      if name.indexOf('pointer') is -1 and not data.listenerNamesRegistered.has(name)
```
jmsmyth: Would `name.indexOf('pointer') isnt 0` suffice? This could catch any events with pointer in the name (if there are any others)

c-frater: Good point, updated

PR #363, opened 2017-01-04
#360 data table sort bugfix

### LINKED ISSUE #360: DataTable: filter ignored when sorting any column in descending order
## Context
Users complain that my data table 'loses' filter when they sort data

## Expected Behavior
Filter applies, only matching rows appear

## Actual Behavior
All rows are visible

## Steps to Reproduce
1. Open https://www.hexagonjs.io/docs/1.9.0/data-table/
2. In the first example type '41' to the filter field (you should see 2 of 3 rows)
3. Click any column 2 times to sort it in descending order

If the demo page works correct, the regression was introduced after version 1.4.2.

## Your Environment
* Hexagon Version: 1.9.0
* Browser Name and version: Chrome 55.0.2883.87 (64-bit) - also reproduced in Firefox 50.1.0
* Operating System and version (desktop or mobile): Kubuntu 16.10

c-frater:
## Description

Updated the object feed to use the same logic for advanced search and regular filters and added handling of the cached filter items.

## Motivation and Context

Resolves #360 

## How Has This Been Tested?

Added regression test and tested in browser

## Checklist:

### DIFF
--- a/modules/data-table/main/index.coffee
+++ b/modules/data-table/main/index.coffee
@@ -1078,13 +1078,14 @@ getAdvancedSearchFilter = (cellValueLookup = hx.identity, termLookup = defaultTe
     hx.defined validFilters
 
 
-getFiltered = (rows, term, filtered, filterCacheTerm, fn) ->
-  if (term?.length and (filtered is undefined or filterCacheTerm isnt term))
+getFiltered = (rows, term, filterCache, filterCacheTerm, fn) ->
+  # term can be a string (regular filter) or an array (advanced search)
+  if (term?.length and (filterCache is undefined or filterCacheTerm isnt term))
     rows.filter fn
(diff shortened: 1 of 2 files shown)


modules/data-table/main/index.coffee
```diff
@@ -1077,6 +1077,15 @@ getAdvancedSearchFilter = (cellValueLookup = hx.identity, termLookup = defaultTe
       not hx.defined invalidFilter
     hx.defined validFilters
 
+
+getFiltered = (rows, term, filtered, filterCacheTerm, fn) ->
```
jmsmyth: Comments explaining what each of these arguments are could help

c-frater: Makes sense - I'll do that now

PR #364, opened 2017-01-04
#358 view bugfix

### LINKED ISSUE #358: View: error with nested views using the same selector
## Context

When building a recursive nested structure for a sitemap, I was using the same selector for the view in the top level as in the nested level(s) without re-creating the entire structure every time.

When calling `apply` the first time, the structure is built correctly.

Subsequent calls then populate nested content with items from the parent level.

## Expected Behavior

Nested views using the same class should not affect or be affected by parent views.

## Actual Behavior

Nested views are populated correctly the first time in but subsequent calls cause the structure to be changed.

## Possible Fix

Use `shallowSelectAll` instead of `selectAll` inside the view

## Steps to Reproduce

Complete example:
```js
const hx = window.hx

function enter () {
  const sel = hx.div('child')

  const children = createView()

  const text = hx.span()

  sel.style('border', '1px solid blue')
    .add(text)
    .add(hx.div('hx-group hx-horizontal')
      .add(hx.section.fixed())
      .add(children))

  sel.api({
    update: (item) => {
      text.text(item.text)
      children.api().update(item.children || [])
    }
  })
  return this.append(sel).node()
}

function update (item, elem) {
  hx.select(elem).api().update(item)
}

function createView () {
  const sel = hx.div('container hx-section')

  const viewDiv = hx.div('view')
  const view = viewDiv.view('.child')
    .enter(enter)
    .update(update)

  sel.api({
    update: (items) => {
      view.apply(items)
    }
  })

  return sel
    .add(viewDiv)
}

const view = createView()

const items = [
  {
    text: 'Item 1',
    children: [
      {
        text: 'Replaced with item 2 on second call'
      }
    ]
  },
  {
    text: 'Item 2'
  }
]

function updateView () {
  view.api().update(items)
}

const updateButton = hx.div('hx-btn hx-action').text('Apply Items')
  .on('click', updateView)

updateView()

hx.select('body').clear()
  .add(hx.div('hx-group hx-vertical')
    .add(hx.section()
      .add(updateButton))
    .add(view))
```

c-frater:
## Description

Converted to use `shallowSelectAll` in views as they shouldn't affect nested children, added regression tests and also resolved a minor issue with `hx.vendor` where it wasn't correctly checking properties. Also fixed an issue with shallowSelectAll where a node didn't have any children.

## Motivation and Context

Resolves #358 

## How Has This Been Tested?

Added regression tests and tested in browser using code in issue

## Checklist:

### DIFF
--- a/modules/selection/main/index.coffee
+++ b/modules/selection/main/index.coffee
@@ -30,18 +30,21 @@ selectSingle = (selector, node) -> getMethod(node, 'querySelector').call(node, s
 selectAll = (selector, node) -> getMethod(node, 'querySelectorAll').call(node, selector)
 
 shallowSelectSingle = (selector, node) ->
-  matchFn = getMatches(node)
-  for child in node.children
-    if matchFn.call(child, selector)
-      return child
+  if node.children?.length
+    matchFn = getMatches(node)
+    for child in node.children
+      if matchFn.call(child, selector)
+        return child
 
 shallowSelectAll = (selector, node) ->
-  matchFn = getMatches(node)
-  matchingNodes = []
-  for child in node.children
-    if matchFn.call(child, selector)
-      matchingNodes.push child
-  matchingNodes
+  if node.children?.length
+    matchFn = getMatches(node)
+    matchingNodes = []
+    for child in node.children
+      if matchFn.call(child, selector)
+        matchingNodes.push child
+    matchingNodes
+  else []
 
 
 getHexagonElementDataObject = (element, createIfNotExists = true) ->
--- a/modules/util/main/index.coffee
+++ b/modules/util/main/index.coffee
@@ -316,7 +316,7 @@ vendorPrefixes = ["webkit", "ms", "moz", "Moz", "o", "O"]
 hx.vendor = (obj, prop) ->
   if prop of obj then return obj[prop]
   for p in vendorPrefixes
-    if (prefixedProp = p + prop.charAt(0) + prop.slice(1)) of obj
+    if (prefixedProp = p + prop.charAt(0).toUpperCase() + prop.slice(1)) of obj
       return obj[prefixedProp]
 
 hx.identity = (d) -> d
--- a/modules/util/test/spec.coffee
+++ b/modules/util/test/spec.coffee
@@ -527,6 +527,19 @@ describe "Util", ->
     hx.clone(m).entries().should.eql(m.entries())
 
 
+  it 'vendor: should find the property if it exists', ->
+    obj = {
+      prop: 'something'
+    }
+    hx.vendor(obj, 'prop').should.equal('something')
+
+
+  it 'vendor: should find a prefixed property if it exists', ->
+    obj = {
+      webkitProp: 'webkit'
+    }
+    hx.vendor(obj, 'prop').should.equal('webkit')
+
 
   it 'cleanNode: should remove all whitespace nodes', ->
     container = document.createElement('div')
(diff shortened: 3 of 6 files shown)


modules/selection/main/index.coffee
```diff
@@ -30,18 +30,21 @@ selectSingle = (selector, node) -> getMethod(node, 'querySelector').call(node, s
...
+        return child
 
 shallowSelectAll = (selector, node) ->
-  matchFn = getMatches(node)
-  matchingNodes = []
-  for child in node.children
-    if matchFn.call(child, selector)
-      matchingNodes.push child
-  matchingNodes
+  if node.children?.length
```
c-frater: Not quite sure _how_ I got this issue but I managed to get into this function with a node without `children` which then fails. Adding this check prevents this from happening

modules/util/main/index.coffee
```diff
@@ -316,7 +316,7 @@ vendorPrefixes = ["webkit", "ms", "moz", "Moz", "o", "O"]
 hx.vendor = (obj, prop) ->
   if prop of obj then return obj[prop]
   for p in vendorPrefixes
-    if (prefixedProp = p + prop.charAt(0) + prop.slice(1)) of obj
+    if (prefixedProp = p + prop.charAt(0).toUpperCase() + prop.slice(1)) of obj
```
c-frater: It looks like it was always _supposed_ to do this (I can't think of another reason for it to get `charAt(0)` if not to capitalise) - Noticed as I was curious as to what hx.vendor actually did.

Added regression test(s)

modules/util/test/spec.coffee
```diff
@@ -85,6 +86,38 @@ describe "Util", ->
...
+    clock.tick(50)
+    debounce()
+    fn.should.not.have.been.called()
+    clock.tick(50)
+    debounce()
+    fn.should.not.have.been.called()
+    clock.tick(101)
+    fn.should.have.been.called()
+
+  it 'debounce: shold pass through arguments', ->
```
c-frater: Noticed some easy-win tests that could be added so added them

modules/view/main/index.coffee
```diff
@@ -23,7 +23,7 @@ class View
       updateSet = []
       exitSet = []
 
-      nodes = @rootSelection.selectAll @selector
+      nodes = @rootSelection.shallowSelectAll @selector
```
c-frater: The main change of this PR - changing selectAll to shallowSelectAll

modules/view/module.json
```diff
@@ -1,6 +1,7 @@
 {
   "dependencies": [
-    "selection"
+    "selection",
+    "component"
```
c-frater: Not sure how to add 'test dependencies' - I've replicated the error in the issue as a test and it uses `.api()` which is part of `component`

modules/view/test/spec.coffee
```diff
@@ -120,4 +120,67 @@ describe 'hx-view', ->
...
+    selection = hx.detached('div')
+
+    view = selection.view('div')
+      .enter(-> @append('div').add(hx.detached('span')).add(hx.detached('div')).node())
+      .update((d, e) -> hx.select(e).select('span').text(d))
+
+    view.apply(['a','b','c'])
+    selection.selectAll('div').text().should.eql(['a','','b','','c',''])
+    view.apply(['a','b','c'])
+    selection.selectAll('div').text().should.eql(['a','','b','','c',''])
```
c-frater: Added a 'simple' version of the error as well as a replication of the error from the issue

PR #367, opened 2017-01-05
#353 Overlapping x axis labels

### LINKED ISSUE #353: Plot: axis labels should never overlap
## Context

Creating a bar chart with interval labels (1AM - 2AM) for 24 hours causes overlapping labels in most cases as there isn't enough room to display all the text correctly.

## Expected Behavior

Graphs with lots of axis tick values (e.g. bar charts) should not allow overlapping of tick values.

In general, it should never be possible for labels to overlap.

## Actual Behavior

Currently, most graphs handle this using the scale, however for discrete axes we seem to always use _all_ the labels.

georgesimms:
#353 : Overlapping x axis labels

## Description

By default, an x axis label will be hidden if there is a label to the left of it with which it overlaps. The old behavior can be retained (ie all x axis labels can be shown) by setting an axis option.

## Motivation and Context
Resolves #353

## How Has This Been Tested?
Manually - by applying the changes in the demo project and comparing before and after (See screenshot) and by adding a test to the plot module.

Before:
screenshot from 2017-01-05 15-47-43
After:
screenshot from 2017-01-05 15-46-54

## Checklist:


modules/plot/main/coffee/Axis.coffee
```diff
@@ -455,25 +466,30 @@ class Axis
...
+            @view('.hx-tick-line', 'line')
+              .update (t) -> @attr('y1', 0).attr('y2', tickSize)
+              .apply(this)
+            @view('.hx-tick-text-x', 'text')
+              .update (t) -> @text(if (i%self.x.nthTickVisible())==0 then self.x.formatter()(t) else '')
+              .apply(tick[0])
+          .apply(if self.x.showTicks() then scale.ticks(self.x.tickSpacing()) else [])
+
+        if self.x.doCollisionDetection()
+          doCollisionDetection @selectAll('.hx-tick-text-x').filter((x) -> x.text()).nodes
```
georgesimms: Among other things, we don't bother doing collision detection for labels that have already been removed by nthTickVisible anyway

modules/plot/main/coffee/Axis.coffee
```diff
@@ -31,6 +41,7 @@ dimension = (axis, options) ->
         state[name]
 
   {
+    doCollisionDetection: setterGetter('doCollisionDetection'),
```
georgesimms: Charlie suggested the user could still need to see all labels (eg if they are angled).

But we do not want to use angled graph labels for other reasons.

modules/plot/main/coffee/Axis.coffee
```diff
@@ -1,12 +1,22 @@
 
 # by far the most complicated component of the graphing api. The auto axis scaling makes things really complicated here...
+doCollisionDetection = (nodes) ->
+  nodes.forEach (node, i) ->
+    previousNodes = nodes.slice 0, i
+    doesOverlap = hx.find previousNodes, (previousNode) ->
```
georgesimms: hx.find should fail early

c-frater:
Changing your example to use different width x values for the bar shows an issue with consistent spacing:

screen shot 2017-01-06 at 07 33 17

X used:
```
x: type === 'line' ? i : `${(i % 4 === 0 ? 'SM' : 'REALLYLONGSTRING')} ${i}`
```

As you can see, the spacing between `SM ${i}` and `REALLYLONGSTRING ${i}` makes them look like the same label

There needs to be the same spacing between labels in terms of ticks, rather than just hiding labels until doesn't overlap.

I'd expect the logic to be effectively equivalent to incrementing the `nthTickVisible` until all the labels fit, otherwise you're left with a very confusing axis

modules/plot/main/coffee/Utils.coffee
```diff
@@ -1,3 +1,19 @@
...
+      currBox = node.getBoundingClientRect()
+      doesCollide = currBox.left < prevBox.right
+      currentDistance = i - j
+      if doesCollide and currentDistance > distance
+        distance = currentDistance
+      doesCollide
+  nodes.forEach (node, i) ->
+    if i % (distance + 1)
+      hx.select node
+        .text ''
```
c-frater: Would `hx.select(node).remove()` be better or does that break the view?

modules/plot/main/coffee/Utils.coffee
```diff
@@ -1,3 +1,19 @@
+doCollisionDetection = (nodes) ->
+  distance = 1
+  nodes.forEach (node, i) ->
+    previousNodes = nodes.slice 0, (i - distance + 1)
+    doesOverlap = previousNodes.find (previousNode, j) ->
+      prevBox = previousNode.getBoundingClientRect()
+      currBox = node.getBoundingClientRect()
+      doesCollide = currBox.left < prevBox.right
+      currentDistance = i - j
```
c-frater: Could we name `i` and `j` a bit better? (nodeIndex and previousIndex or something along those lines?)

modules/plot/main/coffee/Utils.coffee
```diff
@@ -1,3 +1,19 @@
+doCollisionDetection = (nodes) ->
+  distance = 1
+  nodes.forEach (node, i) ->
```
c-frater: Is there a way to make this more functional? (e.g. `distance = resultOfFunction` instead of defining it outside the loop and updating)

Maybe something like this (I haven't tested it at all)
```coffee
doCollisionDetection = (nodes) ->
  distance = hx.max nodes.map (node, nodeIndex) ->
    currBox = node.getBoundingClientRect()
    previousNodes = nodes.slice 0, (nodeIndex - distance + 1)
    hx.max previousNodes.map (previousNode, prevIndex) ->
      prevBox = previousNode.getBoundingClientRect()
      doesCollide = currBox.left < prevBox.right
      if doesCollide then nodeIndex - prevIndex else 1
  nodes.forEach (node, i) ->
    if i % (distance + 1)
      hx.select(node).text ''
```

georgesimms: I think this is done now

modules/plot/main/coffee/Utils.coffee
georgesimms: done

modules/plot/main/coffee/Utils.coffee
georgesimms: I would prefer to leave it like this - nthTickVisible doesn't remove the label, so I don't think we should do so here either.

modules/plot/test/spec.coffee
```diff
@@ -10,6 +10,29 @@ describe "plot", ->
...
+      it 'doCollisionDetection: should hide text that gets in the way of other text', ->
+        widthPx = 200
+        intendedDistance = 2
+        createSuspiciousElement = (i) ->
+          hx.detached 'div'
+            .text i
+            .style 'left', "#{i * widthPx / intendedDistance}px"
+            .style 'width', "#{widthPx}px"
+            .style 'position', 'absolute'
+            .node()
```
georgesimms: new test added. I think I've made it reasonably clear why the distance is what it is.

modules/plot/main/coffee/Utils.coffee
```diff
@@ -1,3 +1,21 @@
+doCollisionDetection = (nodes) ->
+  reductor = (oldDistance, node, currentIndex) ->
+    previousNodes = nodes.slice 0, (currentIndex - oldDistance + 1)
+      .map (previousNode, previousIndex) -> {previousNode, previousIndex}
+    tuple = possibleNewDistance = hx.find previousNodes, ({previousNode, previousIndex}) ->
+      prevBox = previousNode.getBoundingClientRect()
+      currBox = node.getBoundingClientRect()
+      currBox.left < prevBox.right
+    if tuple
+      { previousIndex } = tuple
```
c-frater: We could do this where we call hx.find?

```
{ previousIndex } = hx.find previousNodes, ...
```

In theory, we then can return `hx.max(currentIndex - previousIndex, oldDistance)` and can get rid of this if statement alltogether

modules/plot/main/coffee/Utils.coffee
```diff
@@ -1,3 +1,21 @@
+doCollisionDetection = (nodes) ->
+  reductor = (oldDistance, node, currentIndex) ->
+    previousNodes = nodes.slice 0, (currentIndex - oldDistance + 1)
+      .map (previousNode, previousIndex) -> {previousNode, previousIndex}
+    tuple = possibleNewDistance = hx.find previousNodes, ({previousNode, previousIndex}) ->
+      prevBox = previousNode.getBoundingClientRect()
+      currBox = node.getBoundingClientRect()
```
georgesimms: that's a good point, actually

georgesimms: fixed

modules/plot/main/coffee/Axis.coffee
c-frater: Should probably check `self.x.showTicks()` and the new option value in the if statement.

In the case that showTicks is off currently, we'd effectively call `doCollsionDetection([])` after selecting nothing

c-frater:
Added documentation

PR #371, opened 2017-01-11
#370: Resolves y bounds calculation

### LINKED ISSUE #370: Plot: automatic calculation of ymax does not work when there are multiple series and the data is sparse
## Context
The following stacked bar chart:

```js
const hx = window.hx

const graphSel = hx.graph({
  axes: [{
    x: {
      scaleType: 'discrete'
    },
    y: {
      min: 0
    },
    series: [
      {
        type: 'bar',
        options: {
          title: 'Bob',
          group: 'g0',
          fillColor: hx.theme.plot.coldCol,
          data: [{
            x: 'verybig',
            y: 10
          }, {
            x: 'alwayspresent',
            y: 1
          }]
        }
      },
      {
        type: 'bar',
        options: {
          title: 'Lazlo',
          fillColor: hx.theme.plot.warmCol,
          group: 'g0',
          data: [{
            x: 'alwayspresent',
            y: 2
          }]
        }
      }
    ]
  }]
})

hx.select('#root').add(hx.detached('div').class('hx-content').add(graphSel))
```

## Expected Behavior
The calculated y max value should be 10

## Actual Behavior
The calculated y max value is actually 3

## Possible Fix
The current y max value uses the "top series" to work out what all the x values are for the whole graph. It should get the x values for all series.

## Steps to Reproduce
Use the code above, it results in this screenshot:

screenshot from 2017-01-11 11-11-13

## Your Environment

* Hexagon Version: 1.9.0
* Browser Name and version: Chromium 55.0.2883.87 Built on Ubuntu , running on Ubuntu 16.04
* Operating System and version (desktop or mobile): Ubuntu (desktop) 16.04

georgesimms:
## Description
Y bounds calculation pulled out into its own method

Refactored to make it easier to follow (use Array::map, Array::reduce etc rather than coffeescript for loops)

Y bounds calculation not called when overridden by user manual config

Rather than using topseries, all the x values are got out of all the series

## Motivation and Context
Resolves #370 

## How Has This Been Tested?
Regression test added.
Tested visually with same example as in issue

Screenshot for example in issue:

screenshot from 2017-01-11 14-01-50

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.
- [X] All my changes are covered by tests.
- [ ] This request is ready to merge

### DIFF
--- a/modules/plot/main/coffee/Axis.coffee
+++ b/modules/plot/main/coffee/Axis.coffee
@@ -318,20 +318,9 @@ class Axis
       @yScale.domain(domain)
     else
 
-      { yMinMightBeAuto, yMaxMightBeAuto } = { yMaxMightBeAuto: @y.max(), yMinMightBeAuto: @y.min() }
-      { ymin, ymax } = if 'auto' in [yMaxMightBeAuto, yMinMightBeAuto]
-        { ymin, ymax } = @calculateYBounds()
-        yminscaled = if yMinMightBeAuto is 'auto' then scalePad(ymin, ymax - ymin, -@y.scalePaddingMin()) else yMinMightBeAuto
-        ymaxscaled = if yMaxMightBeAuto is 'auto' then scalePad(ymax, ymax - ymin, @y.scalePaddingMax()) else yMaxMightBeAuto
-        { ymin: yminscaled, ymax: ymaxscaled }
-      else
-        { ymin: yMinMightBeAuto, ymax: yMaxMightBeAuto }
-
-
-      ymin = if @y.min() is 'auto' then ymin else @y.min()
-      ymax = if @y.max() is 'auto' then ymax else @y.max()
-
-
+      yMinMightBeAuto = @y.min()
+      yMaxMightBeAuto = @y.max()
+      { ymin, ymax } = @calculateYBounds yMinMightBeAuto, yMaxMightBeAuto
 
       @yScale.domain(ymin, ymax)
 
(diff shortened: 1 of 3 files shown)


modules/plot/main/coffee/Axis.coffee
```diff
@@ -321,60 +318,20 @@ class Axis
...
     else
 
-      # OPTIM: this expensive calculation is not needed if the scales are non auto
-      # XXX: band series?
+      { yMinMightBeAuto, yMaxMightBeAuto } = { yMaxMightBeAuto: @y.max(), yMinMightBeAuto: @y.min() }
+      { ymin, ymax } = if 'auto' in [yMaxMightBeAuto, yMinMightBeAuto]
+        { ymin, ymax } = @calculateYBounds()
+        yminscaled = if yMinMightBeAuto is 'auto' then scalePad(ymin, ymax - ymin, -@y.scalePaddingMin()) else yMinMightBeAuto
+        ymaxscaled = if yMaxMightBeAuto is 'auto' then scalePad(ymax, ymax - ymin, @y.scalePaddingMax()) else yMaxMightBeAuto
+        { ymin: yminscaled, ymax: ymaxscaled }
```
georgesimms: only run this if either min or max are 'auto'

modules/plot/main/coffee/Axis.coffee
```diff
@@ -321,60 +318,20 @@ class Axis
...
+      else
+        { ymin: yMinMightBeAuto, ymax: yMaxMightBeAuto }
 
-      ymin = undefined
```
georgesimms: code moved into its own method and refactored to use Array::reduce

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,67 @@ class Axis
...
-      for series, j in @series()
+      yStack = Math.max(yscaledomainmin, 0)
+      maybeys = allSeries.map (series) ->
         if series._.seriesId < seriesId and series.group() == group and series._.type == type
-          ys = series.getY(x, @x.scaleType() is 'discrete')
-          if hx.defined(ys)
-            yStack += ys
-      yStack
-    else Math.max(start, 0)
+          series.getY(x, xscaletype is 'discrete')
```
georgesimms: There is a slightly interesting pattern here I've noticed a couple of times where we map and then filter out undefined, to approximate something like scala's collect . I think labeling it like this makes it clearer what's going on

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,67 @@ class Axis
...
+      yStack + hx.sum maybeys.filter hx.identity
+    else
+      Math.max(start, 0)
+
+  calculateYBounds: ->
+    # The series must already be tagged before calling this method
+    # XXX: band series?
```
georgesimms: This comment was there before, I'm not sure if it still applies as we explicitly handle BandSeries below

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,67 @@ class Axis
...
+            extent(data, (d) -> d.y)
+        ys = maybeys.filter((d) -> d?)
+  
+        {
+          yymin: hx.min(ys.map((d) -> d[0]))
+          yymax: hx.max(ys.map((d) -> d[1]))
+        }
+      else
+        topSeries = series[series.length-1]
+        allX = hx.unique hx.flatten series.map (s) -> s.data().map ({ x }) -> x
```
georgesimms: get all the x, not just from the top series

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,69 @@ class Axis
...
+            if s instanceof StraightLineSeries
+              if not data.dx and not data.dy and data.y
+                [data.y, data.y]
+              else
+                undefined
+            else if s instanceof BandSeries
+              extent2(data, ((d) -> d.y1), (d) -> d.y2)
+            else
+              extent(data, (d) -> d.y)
+          ys = maybeys.filter((d) -> d?)
```
georgesimms: Another example of the "map and then filter" pattern

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,69 @@ class Axis
...
+          {
+            yymin: hx.min stackHeights
+            yymax: hx.max stackHeights
+          }
+        {
+          ymin: Math.min(ymin, yymin)
+          ymax: Math.max(ymax, yymax)
+        }
+  
+    stackGroupReductor = (prev, { type, group }) ->
```
georgesimms: we were not actually using the type before. We were pulling it out, but then assuming the topSeries type was the type for all series. So let's just use it directly

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,69 @@ class Axis
...
+    initValue = { ymin: 0, ymax: 0 }
+    types = hx.groupBy(allSeries, (d) -> d._.type)
+    stackGroups = types.map ([type, series]) ->
+      type: type
+      group: hx.groupBy series, (s) -> if supportsGroup(s) then s.group() else undefined
+  
+  
+    typeGroupReductor = (type) =>
+      ({ ymin, ymax }, [seriesGroup, series]) =>
+        { yymin, yymax } = if seriesGroup == undefined
```
georgesimms: again previously we were not actually using seriesGroup beyond checking if it was defined or not

now we use the value we already calculated rather than calculating one from topSeries and assuming it is representative

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,69 @@ class Axis
...
+          ys = maybeys.filter((d) -> d?)
+  
+          {
+            yymin: hx.min(ys.map((d) -> d[0]))
+            yymax: hx.max(ys.map((d) -> d[1]))
+          }
+        else
+          allX = hx.unique hx.flatten series.map (s) -> s.data().map ({ x }) -> x
+          stackHeights = allX.map (x) ->
+            maybeys = series.map (series) -> series.getY x, xscaletype is 'discrete'
```
georgesimms: There is not any point calling getYStack here as it pointlessly just filters over all the series in the graph

we already have the relevant series, so don't waste time filtering them again

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,69 @@ class Axis
...
+            hx.sum maybeys.filter hx.identity
+          {
+            yymin: hx.min stackHeights
+            yymax: hx.max stackHeights
+          }
+        {
+          ymin: Math.min(ymin, yymin)
+          ymax: Math.max(ymax, yymax)
```
georgesimms: ymax and ymin are always numbers now, so the checks for undefined have been removed

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +530,69 @@ class Axis
       xStack
     else Math.max(start, 0)
 
-  # gets the height of the stack for a particular type, group and x value (the seriesId is the series that you want to get the baseline for)
   getYStack: (type, group, x, seriesId, start = 0) ->
```
georgesimms: this is now only called by series, not within the axis itself

modules/plot/test/spec.coffee
```diff
@@ -10,6 +10,49 @@ describe "plot", ->
...
+                  y: 2
+                }]
+              }
+            }
+          ]
+  		})
+
+        axis.tagSeries()
+        { ymax } = axis.calculateYBounds()
+        ymax.should.eql(10)
```
georgesimms: regression test

modules/plot/main/coffee/Axis.coffee
```diff
@@ -321,60 +318,20 @@ class Axis
...
-                if not data.dx and not data.dy and data.y
-                  [data.y, data.y]
-                else
-                  undefined
-              else if s instanceof BandSeries
-                extent2(data, ((d) -> d.y1), (d) -> d.y2)
-              else
-                extent(data, (d) -> d.y)
-            ys = ys.filter((d) -> d?)
+      { yMinMightBeAuto, yMaxMightBeAuto } = { yMaxMightBeAuto: @y.max(), yMinMightBeAuto: @y.min() }
```
c-frater: Not sure I'm keen on destructuring like this - you may as well do `yMinMightbeAuto = @y.min()`

modules/plot/main/coffee/Axis.coffee
```diff
@@ -321,60 +318,20 @@ class Axis
...
-              ymax = 0
-            else
-              ymax = Math.max(ymax, 0)
-            for d in topSeries.data()
-              stackHeight = @getYStack(topSeries._.type, topSeries.group(), d.x, topSeries._.seriesId+1, @yScale.domainMin)
-              if (ymin == undefined or stackHeight < ymin) then ymin = stackHeight
-              if (ymax == undefined or stackHeight > ymax) then ymax = stackHeight
 
       ymin = if @y.min() is 'auto' then ymin else @y.min()
       ymax = if @y.max() is 'auto' then ymax else @y.max()
```
c-frater: Do we still need these lines? It looks like this logic is duplicated above

modules/plot/main/coffee/Axis.coffee
georgesimms: yes, you are right that it is pointless.

georgesimms: removed

modules/plot/main/coffee/Axis.coffee
georgesimms: You may well be right, I'd like to make a test just to make sure though.

I think the easiest way to tell a non-auto value is not ignored is by moving this logic into calculateYBounds and writing a test for it there.

georgesimms: I have removed it, and added tests for when y.min and y.max aren't auto.

modules/plot/main/coffee/Axis.coffee
```diff
@@ -573,14 +519,73 @@ class Axis
...
-            yStack += ys
-      yStack
-    else Math.max(start, 0)
+          series.getY(x, xScaleType is 'discrete')
+      yStack + hx.sum maybeys.filter hx.identity
+    else
+      Math.max(start, 0)
+
+  calculateYBounds: (yMinMightBeAuto, yMaxMightBeAuto) ->
+    if 'auto' in [yMaxMightBeAuto, yMinMightBeAuto]
```
c-frater: +1 for moving this in here

jmsmyth:
Seems like a bugfix to me, so this could be a patch release?

PR #377, opened 2017-01-20
Offer improvement. Remove two useless expressions.

gonaumov:
## Description
There is the following errors into FireFox Web console 
```
TypeError: useless expression hexagon.js:8987:4
TypeError: useless expression hexagon.js:15313:11
```
## Motivation and Context
In fact this code is redundand. I want to help code to be clean and error free. 
## How Has This Been Tested?
I run the tests. I also replace content of hexagon.js file on 
https://www.hexagonjs.io/ with TamperMonkey user script.

## Checklist:

### DIFF
--- a/dist/hexagon-dark/hexagon.js
+++ b/dist/hexagon-dark/hexagon.js
@@ -8985,7 +8985,6 @@ AutoComplete = (function(superClass) {
       _.menu = menu;
       _.input = input;
     }
-    this;
   }
 
   AutoComplete.prototype.clearCache = function() {
@@ -15257,7 +15256,7 @@ ColorPicker = (function(superClass) {
           return circleMoved(pos);
         };
         dragObject = function(elem, parent, min, max, startCallback, moveCallback, endCallback) {
-          var StartListening, StopListening, cursorStartPos, disposed, dragGo, dragStart, dragStop, dragStopHook, dragging, elementStartPos, listening, temp;
+          var StartListening, StopListening, cursorStartPos, dispose, disposed, dragGo, dragStart, dragStop, dragStopHook, dragging, elementStartPos, listening, temp;
           cursorStartPos = elementStartPos = dragging = listening = disposed = null;
           if (min !== null && max !== null) {
             temp = min.min(max);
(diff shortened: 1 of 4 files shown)


modules/color-picker/main/index.coffee
```diff
@@ -227,7 +227,7 @@ class ColorPicker extends hx.EventEmitter
           if endCallback isnt null then endCallback(elem)
           dragging = false
 
-        dispose: ->
+        dispose = ->
```
jmsmyth: It looks like `dispose()` and `disposed` are not used any more so this function can possibly be removed entirely

dist/hexagon-dark/hexagon.js
```diff
@@ -15257,7 +15256,7 @@ ColorPicker = (function(superClass) {
           return circleMoved(pos);
         };
         dragObject = function(elem, parent, min, max, startCallback, moveCallback, endCallback) {
-          var StartListening, StopListening, cursorStartPos, disposed, dragGo, dragStart, dragStop, dragStopHook, dragging, elementStartPos, listening, temp;
```
jmsmyth: The `dist/` files don't normally get updated until a release - could these bits be left out the PR?

modules/color-picker/main/index.coffee
gonaumov: It seems that disposed is used on 8 places:
```
georginaumov@georgi:~/Documents/hexagonjs$ grep -R disposed --include '*.coffee' *
modules/color-picker/main/index.coffee:        cursorStartPos = elementStartPos = dragging = listening = disposed = null
modules/color-picker/main/index.coffee:          if dragging or not listening or disposed then return
modules/color-picker/main/index.coffee:          if !dragging or disposed then return
modules/color-picker/main/index.coffee:          if not dragging or disposed then return true
modules/color-picker/main/index.coffee:          if disposed then return
modules/color-picker/main/index.coffee:          disposed = true
modules/color-picker/main/index.coffee:          if listening or disposed then return
modules/color-picker/main/index.coffee:          if not listening or disposed then return
georginaumov@georgi:~/Documents/hexagonjs$ grep -R disposed --include '*.coffee' * | wc -l
8
```
but I can remove it. Are you sure? By the way this is my first try with cofeescript so may be I missing something.

jmsmyth: Should be ok I think. The only place `disposed` is assigned to is in the `dispose` function which isn't used, so `disposed` is always going to be null.

gonaumov: This is done. I removed dispose method and disposed variable completely.

dist/hexagon-dark/hexagon.js
gonaumov: This is done.

Issue #378, opened 2017-01-20
HexagonJS doesn't support timezones.


gonaumov:
When I worked on this PR:
https://github.com/ocadotechnology/hexagonjs/pull/377
I run the tests and first I see the following errors:
```
✖ 5 tests failed

FAILED TESTS:
  date-localizer
    dateTimeLocalizer
      ✖ time: should localize a date object to return a time string of hh:mm
        PhantomJS 1.9.8 (Linux 0.0.0)
      AssertionError: expected '3:30' to equal '1:30'
          at /home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:210
          at assertEqual (/home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:782)
          at /home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:4192
          at /home/georginaumov/Documents/hexagonjs/target/test-library/lib/date-localizer/spec.js:66

      ✖ time: should localize a date object to return a time string of hh:mm:ss
        PhantomJS 1.9.8 (Linux 0.0.0)
      AssertionError: expected '3:30:00' to equal '1:30:00'
          at /home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:210
          at assertEqual (/home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:782)
          at /home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:4192
          at /home/georginaumov/Documents/hexagonjs/target/test-library/lib/date-localizer/spec.js:69

      ✖ stringToDate: should convert a localized date string back to a date object
        PhantomJS 1.9.8 (Linux 0.0.0)
      AssertionError: expected Wed, 06 Jan 2016 22:00:00 GMT to deeply equal Thu, 07 Jan 2016 00:00:00 GMT
          at /home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:210
          at assertEql (/home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:815)
          at /home/georginaumov/Documents/hexagonjs/node_modules/chai/chai.js:4192
          at /home/georginaumov/Documents/hexagonjs/target/test-library/lib/date-localizer/spec.js:79

      ✖ stringToDate: should convert a localized date string back to a date object for inbuilt dates
        PhantomJS 1.9.8 (Linux 0.0.0)
[...]

PR #379, opened 2017-01-21
WIP:Resolved issue with timezones different by UTC:0.00.

gonaumov:
## Description
I found this issue:
https://github.com/ocadotechnology/hexagonjs/issues/378
when I worked on this PR:
https://github.com/ocadotechnology/hexagonjs/pull/377 
The issue can be reproduced on current master branch when you set timezone 
different from UTC:0.00. For example on the master branch try to set your location to Sofia/Bulgaria (UTC: +2) and run tests, after this try to set your location back to UTC: 0.00  - London - in this case npm test will pass.  Please review it carefully because with issues like that
sometimes it is very hard to decide where is the problem - in the method or 
in the test.  

## Motivation and Context
This change is required because HexagonJS should work properly on all timezones.
## How Has This Been Tested?
I run the tests several times with different timezones. 
## Checklist:

### DIFF
--- a/modules/date-localizer/main/index.coffee
+++ b/modules/date-localizer/main/index.coffee
@@ -25,23 +25,20 @@ class DateTimeLocalizer
   year: (year) -> year
 
   # localise a date object to return a date string of dd/mm/yyyy (or localised format)
-  date: (inputDate, useInbuilt) ->
-    # We should clone date not change
-    # agruments - according to tests.
-    resultDate = new Date(inputDate.getTime())
+  date: (date, useInbuilt) ->
     if useInbuilt
-      resultDate.getFullYear() + '-' +
-      zeroPad(resultDate.getMonth() + 1 )+ '-' +
-      zeroPad(resultDate.getDate())
+      date.getFullYear() + '-' +
+      zeroPad(date.getMonth() + 1 )+ '-' +
+      zeroPad(date.getDate())
     else
-      zeroPad(resultDate.getDate()) + '/' + zeroPad(resultDate.getMonth() + 1) + '/' + resultDate.getFullYear()
+      zeroPad(date.getDate()) + '/' + zeroPad(date.getMonth() + 1) + '/' + date.getFullYear()
 
   # localise a date object to return a time string of hh:mm or hh:mm:ss (or localised format)
-  time: (inputDate, enableShowSeconds) ->
-    resultDate = hx.preferences.applyTimezoneOffset(inputDate)
-    timeString = '' + zeroPad resultDate.getHours() + ':' + zeroPad resultDate.getMinutes()
-    if enableShowSeconds
-      timeString = timeString.concat ':' + zeroPad resultDate.getSeconds()
+  time: (date, showSeconds) ->
+    date = hx.preferences.applyTimezoneOffset(date)
+    timeString = date.getHours() + ':' + zeroPad date.getMinutes()
+    if showSeconds
+      timeString += ':' + zeroPad date.getSeconds()
     timeString
 
   # check if a time is a valid time (time as array of [hh, mm, ss])
--- a/modules/date-localizer/main/index.coffee
+++ b/modules/date-localizer/main/index.coffee
@@ -70,7 +70,7 @@ class DateTimeLocalizer
             year = Number split[i]
             if year.toString().length is 2 then year += 2000
       if daysValid and monthsValid and yearsValid
-        new Date(year, month - 1, day)
+        new Date(Date.UTC(year, month - 1, day))
       else
         new Date('Invalid Date')
     else
--- a/modules/date-localizer/main/index.coffee
+++ b/modules/date-localizer/main/index.coffee
@@ -25,20 +25,23 @@ class DateTimeLocalizer
   year: (year) -> year
 
   # localise a date object to return a date string of dd/mm/yyyy (or localised format)
-  date: (date, useInbuilt) ->
+  date: (inputDate, useInbuilt) ->
+    # We should clone date not change
+    # agruments - according to tests.
+    resultDate = new Date(inputDate.getTime())
     if useInbuilt
-      date.getFullYear() + '-' +
-      zeroPad(date.getMonth() + 1 )+ '-' +
-      zeroPad(date.getDate())
+      resultDate.getFullYear() + '-' +
+      zeroPad(resultDate.getMonth() + 1 )+ '-' +
+      zeroPad(resultDate.getDate())
     else
-      zeroPad(date.getDate()) + '/' + zeroPad(date.getMonth() + 1) + '/' + date.getFullYear()
+      zeroPad(resultDate.getDate()) + '/' + zeroPad(resultDate.getMonth() + 1) + '/' + resultDate.getFullYear()
 
   # localise a date object to return a time string of hh:mm or hh:mm:ss (or localised format)
-  time: (date, showSeconds) ->
-    date = hx.preferences.applyTimezoneOffset(date)
-    timeString = date.getHours() + ':' + zeroPad date.getMinutes()
-    if showSeconds
-      timeString += ':' + zeroPad date.getSeconds()
+  time: (inputDate, enableShowSeconds) ->
+    resultDate = hx.preferences.applyTimezoneOffset(inputDate)
+    timeString = '' + zeroPad resultDate.getHours() + ':' + zeroPad resultDate.getMinutes()
+    if enableShowSeconds
+      timeString = timeString.concat ':' + zeroPad resultDate.getSeconds()
     timeString
 
   # check if a time is a valid time (time as array of [hh, mm, ss])
--- a/modules/date-localizer/test/spec.coffee
+++ b/modules/date-localizer/test/spec.coffee
@@ -4,27 +4,27 @@ describe 'dateTimeLocalizerMoment', ->
 describe 'dateTimeLocalizer', ->
   localizer = hx.dateTimeLocalizer()
   zeroPad = hx.format.zeroPad(2)
+  timezonesOffsetPattern = /(.)(\d{2}):(\d{2})$/
   testDate = new Date(1452130200000) # Thu Jan 07 2016 01:30:00 GMT+0000 (GMT)
   zeroHourDate = new Date(1452124800000) # Thu Jan 07 2016 00:00:00 GMT+0000 (GMT)
   invalidDate = new Date('Invalid Date')
   # 0 - London, -120 - Sofia, -60 Krakow, -540 Tokio, 420 Pinal (Arizona/United states)
-  timezonesOfsets = [0, -120, -60, -540, 420]
+  timezonesOffsets = [0, -120, -60, -540, 420]
 
-  getHexaGonTimeZoneOffset = ->
-    hexagonTimeZoneOffset = hx.preferences.timezone().replace(/^UTC\+?/,'').split(":").map (item, index) ->
-      if index is 0
-        Number(item) * 60
-      else
-        Number(item)
-    .reduce (current, previous) ->
-      current + previous
-    hexagonTimeZoneOffset
+  getHexagonTimeZoneOffset = ->
+    timezonesOffsetParts = timezonesOffsetPattern.exec(hx.preferences.timezone())
+    if timezonesOffsetParts
+      [..., sign, hours, minutes] = timezonesOffsetParts;
+      (Number(sign + hours) * 60) + Number(minutes)
+    else
+      hx.consoleWarning('Cannot calculate Hexagon internal timezone offset')
+      false
 
   # This is helper function. It accepts
   # function as a argument and execute it
   # in all timezones that is provided
   executeFunctionInAllTimeZones = (func) ->
-    for currentOffset in timezonesOfsets
+    for currentOffset in timezonesOffsets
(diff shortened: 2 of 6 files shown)


gonaumov:
There is a need for an additional debugging, investigating and more unit tests like this: 
```
  it 'time: should localize a date object to return a time string of hh:mm in Sofia timezone', ->
    stub = sinon.stub(Date.prototype, 'getTimezoneOffset', () -> -120)
    localizer.time(testDate()).should.equal('1:30')
    stub.restore()
```
for several timezones. Will cancel it and will create new when I'm 100% sure where the problem is.

modules/preferences/main/index.coffee
```diff
@@ -284,10 +284,8 @@ class Preferences extends hx.EventEmitter
   supportedTimezones: option 'supportedTimezones'
   timezoneOffsetLookup: option 'timezoneOffsetLookup'
 
-  applyTimezoneOffset: (date, offset) ->
-    offset ?= @_.timezoneOffsetLookup(@timezone(), date.getTime()) || 0
-    utc = date.getTime() + (date.getTimezoneOffset() * 60000)
-    new Date(utc + offset * 60 * 60 * 1000)
+  applyTimezoneOffset: (date) ->
+    new Date(date.getTime() + (date.getTimezoneOffset() * 60000))
```
gonaumov: In fact the previous version worked properly only when offset was equal to 0 - on Greenwich time zone. Try to run the tests and to change your time-zone manually.

gonaumov:
I changed the code in two places and implement several unit test that tests correct behaviour.

modules/preferences/main/index.coffee
jmsmyth: The behaviour of the api seems to have changed a bit here - it's no longer using the timezone from the preferences and no longer lets you optionally pass in a different offset.

I wonder if `executeFunctionInAllTimeZones` just needs updating to set up `preferences.timezone` and `preferences.timezoneOffsetLookup`?

gonaumov: I'll try to do it this night. The problem is that when you change the time zone of your computer tests really do not work.

gonaumov: @jmsmyth I think I found where the problem was. Please look at it again.

modules/preferences/main/index.coffee
```diff
@@ -287,7 +287,13 @@ class Preferences extends hx.EventEmitter
   applyTimezoneOffset: (date, offset) ->
     offset ?= @_.timezoneOffsetLookup(@timezone(), date.getTime()) || 0
     utc = date.getTime() + (date.getTimezoneOffset() * 60000)
-    new Date(utc + offset * 60 * 60 * 1000)
+    # In fact timezoneOffsetLookup working properly and returns
+    # 2 for Sofia but the proper offset is not this because
+    # If we want to get real date we must subtract
+    # these hours - the real offset for Sofia for
+    # example is -120
+    realOffset = -1 * offset * 60
```
gonaumov: If offset is negative the real offset must be positive and vice versa.

jmsmyth: Is the `* 60` needed, since its also multiplying by `60 * 60 * 1000` below?

gonaumov: Try to run the tests. If returned offset was in minutes the below multiplication will be enough.

modules/preferences/main/index.coffee
```diff
@@ -286,8 +286,7 @@ class Preferences extends hx.EventEmitter
 
   applyTimezoneOffset: (date, offset) ->
     offset ?= @_.timezoneOffsetLookup(@timezone(), date.getTime()) || 0
-    utc = date.getTime() + (date.getTimezoneOffset() * 60000)
-    new Date(utc + offset * 60 * 60 * 1000)
+    new Date(date.getTime() + -1 * offset * 60 * 60 * 1000)
```
gonaumov: This is very interesting issue. I wondered why is this magical need for multiple with 60 one time more and in the end I found the problem. **date.getTime()** return the time in UTC. There is no need to subtract the time-zone offset.

modules/preferences/main/index.coffee
gonaumov: You were right about unnecessary extra multiplication with 60.

gonaumov:
@jmsmyth I think the problem is fixed now.  Could you check it again?

modules/preferences/main/index.coffee
c-frater: This has reversed the desired behaviour of the localization:
```js
const date = new Date(1485333362249)
hx.dateTimeLocalizer().time(date)
// Returns -> "8:36"
hx.preferences.timezone("UTC-08:00")
hx.dateTimeLocalizer().time(date)
// Returns -> "16:36"
```

The desired behaviour is to display the current time relative to the timezone so should return `0:36` in this case.

The localization currently mimics that provided by `moment.js`, where we took a lot of the logic from. We also use moment.js under the hood (if it is available in the browser).

Equivalent Moment.js:
```js
const momentDate = moment(1485333362249)
momentDate.format('H:mm')
// Returns -> "8:36"
momentDate.tz('America/Los_Angeles').format('H:mm')
// Returns -> "0:36"
```

As far as I can see, you're correct in removing where we  were using `date.getTimezoneOffset`, I think removing the `-1 *` should resolve the issue where the behaviour is flipped.

Could you also add a few positive values to the `timezoneOffsets`? 

We should probably also test with different `hx.preferences.timezone` options set as we are using them to determine what the time should be.

gonaumov: I added a positive time-zone. Interestingly, when you change the date to this time-zone manually  - I mean the PC system time the tests not pass. I will debug more this evening after work and will try your suggestion.

gonaumov:
@jmsmyth @c-frater I just pushed this new version. This version follows the desired behaviour of the localization and
``` 
npm test
``` 
also passed in different time-zones. I tried to switch my local date to London. Sofia and Pinal - Arizona.

modules/date-localizer/test/spec.coffee
```diff
@@ -3,79 +3,221 @@ describe 'dateTimeLocalizerMoment', ->
...
+      # The atempt below with
+      # hx.preferences.timezone(timeZone)
+      # really work.
+      Date = do (newDate = Date) ->
+        ->
+          dateResult = new (Function.prototype.bind.apply(newDate, arguments))
+          dateResult.setTime(dateResult.getTime() + (currentOffset * 60000))
+          dateResult
+      func()
+      Date = oldDate
```
gonaumov: Actually code from 37nd to 43nd line has no effect under nodejs. In the browser it works perfectly. There is no errors and etc but Date object behaviour remains unchanged. I tried many things like
```
sinon.useFakeTimers calculatedMiliseconds, 'Date'
```
but this also not helped. I left the code here to show the issue. We can remove it. The code bellow - from 44 to 48
```
     oldTimeZone = hx.preferences.timezone()
     timeZone = hx._.preferences.defaultTimezoneLookup(currentOffset)
     hx.preferences.timezone(timeZone)
     func()
     hx.preferences.timezone(oldTimeZone)
```
line working properly but in fact this changes Hexagon internal time-zone not browser time-zone.

modules/date-localizer/test/spec.coffee
```diff
@@ -3,79 +3,221 @@ describe 'dateTimeLocalizerMoment', ->
...
+
+  getHexaGonTimeZoneOffset = ->
+    hexagonTimeZoneOffset = hx.preferences.timezone().replace(/^UTC\+?/,'').split(":").map (item, index) ->
+      if index is 0
+        Number(item) * 60
+      else
+        Number(item)
+    .reduce (current, previous) ->
+      current + previous
+    hexagonTimeZoneOffset
```
gonaumov: This function can also be quite short. I can offer something like this if you prefer: 
```
result = /(.)(\d{2}):(\d{2})$/.exec("UTC-02:00")
if result 
  [sign,hoursPart,minutesPart] = result.slice(1)
  console.log sign, hoursPart, minutesPart
else 
  console.log "Cannot extract date parts."
```

modules/preferences/main/index.coffee
```diff
@@ -284,10 +295,10 @@ class Preferences extends hx.EventEmitter
   supportedTimezones: option 'supportedTimezones'
   timezoneOffsetLookup: option 'timezoneOffsetLookup'
 
-  applyTimezoneOffset: (date, offset) ->
```
jmsmyth: Looking at this, I think the new code you've replaced it with does exactly the same as before. Is the idea just to split it up a bit to make it clearer?

If thats the case, could we make `handleTimeZoneOffsetDifferences` a function rather than a method?

modules/preferences/test/spec.coffee
```diff
@@ -54,7 +54,7 @@ describe 'hx-preferences', ->
 
     describe 'timezone', ->
       it 'setter/getter', ->
-        hx.preferences.timezone('UTC+01:00').should.equal(hx.preferences)
```
jmsmyth: This was doing the correct thing before - most of the apis in hexagon return themselves when setting so that you can chain like this:

```
hx.preferences
  .timezone('UTC+1:00')
  .locale('bg')
```

gonaumov: This is precisely one of the things for which tests are broke under a different timezone from GTM+00:00.
There is need for additional debugging, which I will do in my free time.

modules/date-localizer/test/spec.coffee
c-frater: I prefer this method, although if you could pull out the regex (`/(.)(\d{2}):(\d{2})$/`) to a variable at the top (e.g. next to `zeroPad`) it would be slightly more efficient 👍

modules/date-localizer/test/spec.coffee
```diff
@@ -3,79 +3,221 @@ describe 'dateTimeLocalizerMoment', ->
...
   localizer = hx.dateTimeLocalizer()
-
+  zeroPad = hx.format.zeroPad(2)
   testDate = new Date(1452130200000) # Thu Jan 07 2016 01:30:00 GMT+0000 (GMT)
   zeroHourDate = new Date(1452124800000) # Thu Jan 07 2016 00:00:00 GMT+0000 (GMT)
   invalidDate = new Date('Invalid Date')
+  # 0 - London, -120 - Sofia, -60 Krakow, -540 Tokio, 420 Pinal (Arizona/United states)
+  timezonesOfsets = [0, -120, -60, -540, 420]
+
+  getHexaGonTimeZoneOffset = ->
```
c-frater: Minor thing but could this not have a capital `G` in `Hexagon`? :)

modules/preferences/main/index.coffee
gonaumov: I use handleTimeZoneOffsetDifferences into test also. Like a post-converter.

modules/date-localizer/test/spec.coffee
c-frater: I think testing with both the hexagon timezone and the system timezone changed makes sense because as we've seen the changing of the system timezone can cause issues (e.g. the ones this PR is trying to resolve)

I'm not sure why `sinon.useFakeTimers` doesn't work but I'm fine with the workaround

modules/preferences/main/index.coffee
c-frater: I agree - it should be a separate function

modules/date-localizer/test/spec.coffee
gonaumov: This is done.

modules/preferences/main/index.coffee
gonaumov: There was a problem with applyTimezoneOffset usage so I need to reapply utc time manually. Please check it again. If this solution is not acceptable for you I will search for another.

modules/preferences/test/spec.coffee
gonaumov: Sorry. I removed this change.

modules/date-localizer/main/index.coffee
```diff
@@ -25,20 +25,23 @@ class DateTimeLocalizer
...
   # localise a date object to return a time string of hh:mm or hh:mm:ss (or localised format)
-  time: (date, showSeconds) ->
-    date = hx.preferences.applyTimezoneOffset(date)
-    timeString = date.getHours() + ':' + zeroPad date.getMinutes()
-    if showSeconds
-      timeString += ':' + zeroPad date.getSeconds()
+  time: (inputDate, enableShowSeconds) ->
+    resultDate = hx.preferences.applyTimezoneOffset(inputDate)
+    timeString = '' + zeroPad resultDate.getHours() + ':' + zeroPad resultDate.getMinutes()
+    if enableShowSeconds
```
gonaumov: I removed this change.

modules/date-localizer/test/spec.coffee
gonaumov: This is fixed.

modules/date-localizer/test/spec.coffee
gonaumov: All right. Let this code to stay in place.

gonaumov:
@jmsmyth @c-frater Please review it again. In fact, the code has only one change. Everything else is in the tests. The profit from this PR is that tests pass in all time-zones.

PR #383, opened 2017-01-29
#372: added hx.sort.compareNullsLast

### LINKED ISSUE #372: hx.sort.compare treats `undefined` and `null` as normal strings
## Context

using hx.sort.compare to compare strings with undefined or null values
## Expected Behavior

```js
hx.sort.compare("m", null)
// 1

hx.sort.compare("o", null)
// 1

hx.sort.compare("t", undefined)
// 1

hx.sort.compare("v", undefined)
// 1

hx.sort.compare(null, undefined)
// 0

hx.sort.compare(undefined, null)
// 0
```
I'd mainly expect `undefined` or `null` values to be compared consistently and not according to their string representation. Whether or not the comparison of a defined value with `undefined` or `null` returns -1 or 1 is up for debate.
The last two examples may be debatable as well, if you want to prioritise one over the other.
## Actual Behavior

```js
hx.sort.compare("m", null)
// -1

hx.sort.compare("o", null)
// 1

hx.sort.compare("t", undefined)
// -1 

hx.sort.compare("v", undefined)
// 1

hx.sort.compare(null, undefined)
// -1

hx.sort.compare(undefined, null)
// 1
```
## Possible Fix

Check for whether or not each value is defined in the comparison.
## Steps to Reproduce

1. go to https://www.hexagonjs.io/
2. run above commands in the browser console

## Your Environment

* Hexagon Version: 1.9.0
* Browser Name and version: Chromium 55.0.2883.87, Firefox 50.1.0
* Operating System and version (desktop or mobile): Ubuntu 16.04 (64-bit)

jonthans:
## Description

* made behaviour of custom collator more consistent Intl.Collator
* made usage of Intl.Collator or custom collator changeable in tests
* added hx.sort.compareNullsLast, including tests

## Motivation and Context

Resolves #372

## How Has This Been Tested?

added tests for compareNullsLast and extended existing tests to be run with custom collator and Intl Collator.
Temporarily added Firefox and Chrome as test environments to test Intl Collator behaviour

## Checklist:

### DIFF
--- a/modules/sort/main/index.coffee
+++ b/modules/sort/main/index.coffee
@@ -41,20 +41,16 @@ localeCollatorFn = (locale, options) ->
 # slower than compare but enforces locale comparison for browsers that
 # dont support Intl.Collator.
 localeCompare = (locale, options) ->
-  options ?= {numeric: true}
+  options = hx.merge.defined(options, { numeric: true })
 
   localeCollator = localeCollatorFn(locale, options)
 
-  (a, b) ->
-    if a? and b? and not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
-    else localeCollator(a, b)
-
-localeCompareNullsLast = (locale, options) ->
-  options ?= {numeric: true}
-
-  localeCollator = localeCollatorFn(locale, options)
-  nullsLastCollator(localeCollator)
-
+  if options.nullsLast
+    nullsLastCollator(localeCollator)
+  else
+    (a, b) ->
+      if a? and b? and not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
+      else localeCollator(a, b)
 
 hx.sortBy = (arr, f) ->
   newArr = [arr...]
@@ -68,4 +64,3 @@ hx.sort = (arr) -> hx.sortBy arr, (x) -> x
 hx.sort.compare = compare
 hx.sort.compareNullsLast = compareNullsLast
 hx.sort.localeCompare = localeCompare
-hx.sort.localeCompareNullsLast = localeCompareNullsLast
--- a/modules/sort/main/index.coffee
+++ b/modules/sort/main/index.coffee
@@ -2,18 +2,25 @@
 # https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator
 # using a collator is supposed to be faster than doing localeCompare
 hasCollator = Intl?.Collator?
+hasCollatorFn = if (window.chai) then -> Intl?.Collator else -> hasCollator
 
-collator = if hasCollator
+collatorFn = -> if hasCollatorFn()
   new Intl.Collator(undefined, {numeric: true}).compare
 else
   (a, b) ->
-    if a < b then -1
-    else if a > b then 1
-    else 0
+    if a is b then 0
+    else if String(a) > String(b) then 1
+    else -1
 
-compare = (a, b) ->
-  if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
-  else collator(a, b)
+compare = if (window.chai)
+    (a, b) ->
+      if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
+      else collatorFn()(a, b)
+  else
+    collator = collatorFn()
+    (a, b) ->
+      if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
+      else collator(a, b)
 
 
 # slower than compare but enforces locale comparison for browsers that
@@ -21,7 +28,7 @@ compare = (a, b) ->
 localeCompare = (locale, options) ->
   options ?= {numeric: true}
 
-  localeCollator = if hasCollator
+  localeCollator = if hasCollatorFn()
     new Intl.Collator(locale, options).compare
   else
     (a, b) ->
--- a/modules/sort/main/index.coffee
+++ b/modules/sort/main/index.coffee
@@ -13,11 +13,6 @@ else
     else if String(a) < String(b) then -1
     else 1
 
-compare = (a, b) ->
-  hx._.sort.collator ?= collatorFn()
-  if a? and b? and not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
-  else hx._.sort.collator(a, b)
-
 nullsLastCollator = (collator) -> (a, b) ->
   if a is b then 0
   else if a is undefined then 1
@@ -27,6 +22,13 @@ nullsLastCollator = (collator) -> (a, b) ->
   else if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
   else collator(a, b)
 
+defaultCollator = (collator) -> (a, b) ->
+  if a? and b? and not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
+  else collator(a, b)
+
+compare = (a, b) ->
+  hx._.sort.collator ?= collatorFn()
+  defaultCollator(hx._.sort.collator)(a, b)
 
 compareNullsLast = (a, b) ->
   hx._.sort.collator ?= collatorFn()
@@ -48,9 +50,7 @@ localeCompare = (locale, options) ->
   if options.nullsLast
     nullsLastCollator(localeCollator)
   else
-    (a, b) ->
-      if a? and b? and not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
-      else localeCollator(a, b)
+    defaultCollator(localeCollator)
 
 hx.sortBy = (arr, f) ->
   newArr = [arr...]
--- a/modules/sort/main/index.coffee
+++ b/modules/sort/main/index.coffee
@@ -1,26 +1,21 @@
 # Intl.Collator isn't supported by safari
 # https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator
 # using a collator is supposed to be faster than doing localeCompare
-hasCollator = Intl?.Collator?
-hasCollatorFn = if (window.chai) then -> Intl?.Collator else -> hasCollator
+hasCollator = -> Intl?.Collator?
 
-collatorFn = -> if hasCollatorFn()
+collatorFn = -> if hasCollator()
   new Intl.Collator(undefined, {numeric: true}).compare
 else
   (a, b) ->
     if a is b then 0
     else if String(a) > String(b) then 1
     else -1
+    
+collator = if (window.chai?) then (a, b) -> collatorFn()(a, b) else collatorFn()
 
-compare = if (window.chai)
-    (a, b) ->
-      if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
-      else collatorFn()(a, b)
-  else
-    collator = collatorFn()
-    (a, b) ->
-      if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
-      else collator(a, b)
+compare = (a, b) ->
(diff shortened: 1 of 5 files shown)


modules/sort/main/index.coffee
```diff
@@ -1,15 +1,17 @@
...
-collator = if hasCollator
+collatorFn = -> if hasCollator()
   new Intl.Collator(undefined, {numeric: true}).compare
 else
   (a, b) ->
-    if a < b then -1
-    else if a > b then 1
-    else 0
+    if a is b then 0
+    else if String(a) > String(b) then 1
```
jonthans: This makes the behaviour of the custom collator more consistent with the behaviour of Intl.Collator, since it also turns `undefined` into 'undefined' and `null` into 'null'.

modules/sort/main/index.coffee
```diff
@@ -1,15 +1,17 @@
...
+    else -1
+    
+collator = if (window.chai?) then (a, b) -> collatorFn()(a, b) else collatorFn()
```
jonthans: This makes the collator check whether or not the Intl.Collator is defined or not each time it is being called, but only in tests

modules/sort/main/index.coffee
```diff
@@ -31,6 +33,13 @@ localeCompare = (locale, options) ->
     if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
     else localeCollator(a, b)
 
+compareNullsLast = (a, b) ->
+  if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
+  else
+    if a is b then 0
+    else if a is undefined or (a is null and b?) then 1
```
jonthans: This sorts `undefined` and `null` to the end while keeping `undefined > null`

modules/sort/test/spec.coffee
```diff
@@ -47,55 +57,78 @@ describe 'hx-sort tests', ->
       byName.map((x) -> x.name).should.eql(['Bob', 'Ganesh', 'Kate', 'Lazlo'])
       byAge.map((x) -> x.name).should.eql(['Lazlo', 'Ganesh', 'Bob', 'Kate'])
 
-
-  describe 'using standard compare. ', ->
+  compareTests = ->
```
jonthans: extracting test blocks to be reused

modules/sort/test/spec.coffee
```diff
@@ -47,55 +57,78 @@ describe 'hx-sort tests', ->
...
 
     it 'Arrays of numbers should return the right value', ->
       array = [100, 200, 1, 20, 2, 10]
-      array.sort(hx.sort.compare).should.eql([1,2,10,20,100,200])
+      array.sort(hx.sort.compare).should.eql([1, 2, 10, 20, 100, 200])
+
+  describe 'using standard compare potentially with Intl.Collator. ', compareTests
 
-  describe 'using localeCompare. ', ->
+  describe 'using standard compare without Intl.Collator', runWithoutCollator(compareTests)
```
jonthans: running the normal `compare` tests (potentially) with Intl.Collator and without.

modules/sort/test/spec.coffee
```diff
@@ -108,3 +141,52 @@ describe 'hx-sort tests', ->
...
+  compareNullsLastTests = ->
+    it 'Strings should return the right value', ->
+      hx.sort.compareNullsLast('a', 'b').should.equal(-1)
+      hx.sort.compareNullsLast('b', 'a').should.equal(1)
+      hx.sort.compareNullsLast('a', 'a').should.equal(0)
+
+    it 'undefined should always be sorted to the end', ->
+      hx.sort.compareNullsLast(undefined, 'a').should.equal(1)
+      hx.sort.compareNullsLast('a', undefined).should.equal(-1)
+      hx.sort.compareNullsLast(undefined, 'z').should.equal(1)
```
jonthans: note that `hx.sort.compareNullsLast(undefined, 'z') === -1` and `hx.sort.compareNullsLast(null, 'z') === -1` in normal compare

modules/sort/main/index.coffee
c-frater: We could get around the need to do this by storing collator as `hx._.sort.collator` and deleting it in the tests.
I'm not keen on adding code to hexagon to support tests...

e.g.
`hx._.sort.compare = collatorFn()`

Use `hx._.sort.compare(a, b)` in `compare`

Then in tests you can do `delete hx._.sort.compare` after removing `Intl`

This way it's agnostic of the way the tests are implemented and still retains the 'singleton' behaviour in the browser

modules/sort/main/index.coffee
c-frater: Could possibly add `localeCompareNullsLast`?

modules/sort/test/spec.coffee
```diff
@@ -108,3 +141,52 @@ describe 'hx-sort tests', ->
...
+    it 'Numbers should return the right value', ->
+      hx.sort.compare(1, 2).should.equal(-1)
+      hx.sort.compare(2, 1).should.equal(1)
+      hx.sort.compare(1, 1).should.equal(0)
+
+    it 'Arrays of numbers should return the right value', ->
+      array = [100, 200, 1, 20, 2, 10]
+      array.sort(hx.sort.compare).should.eql([1, 2, 10, 20, 100, 200])
+
+  describe 'using compareNullsLast potentially with Intl.Collator. ', compareNullsLastTests
```
c-frater: 'potentially' sounds a bit dubious...

modules/sort/test/spec.coffee
c-frater: Makes sense - as long as this behaviour matches `Intl.collator(...).compare`
N.B. when using `Array.sort` the `undefined` values will never be passed to `hx.sort.compare` and will always be at the end of the resulting array

modules/sort/test/spec.coffee
c-frater: Why 'potentially'?

modules/sort/test/spec.coffee
c-frater: Test could just be `using compareNullsLast` and `using compareNullsLast without Intl.Collator`

modules/sort/test/spec.coffee
```diff
@@ -108,3 +141,52 @@ describe 'hx-sort tests', ->
...
+
+    it 'Arrays should return the right value', ->
+      array = ['c', 'a', 'b', 'aa', 'ab', 'ac', '1', '2']
+      array.sort(hx.sort.compareNullsLast).should.eql(['1', '2', 'a', 'aa', 'ab', 'ac', 'b', 'c'])
+      array = ['25', '10', '20', '42', '100']
+      array.sort(hx.sort.compareNullsLast).should.eql(['10', '20', '25', '42', '100'])
+
+    it 'Arrays of strings with undefined values should sort null and undefined to the end', ->
+      array = ['b', undefined, null, 'z', 'a', 'p']
+      array.sort(hx.sort.compareNullsLast).should.eql(['a', 'b', 'p', 'z', null, undefined])
```
c-frater: could you add multiple `undefined` and `null` to the array? Just for sanity check on `if a is b`

modules/sort/test/spec.coffee
```diff
@@ -108,3 +141,52 @@ describe 'hx-sort tests', ->
...
+
+    it 'Numbers should return the right value', ->
+      hx.sort.compare(1, 2).should.equal(-1)
+      hx.sort.compare(2, 1).should.equal(1)
+      hx.sort.compare(1, 1).should.equal(0)
+
+    it 'Arrays of numbers should return the right value', ->
+      array = [100, 200, 1, 20, 2, 10]
+      array.sort(hx.sort.compare).should.eql([1, 2, 10, 20, 100, 200])
```
c-frater: Add `null` here (or an additional test with null)

Also should this be `hx.sort.compareNullsLast`?

modules/sort/main/index.coffee
jonthans: That's a really good idea. I'll look into that

modules/sort/test/spec.coffee
jonthans: E.g. Phantom doesn't have Intl.Collator (which is why I've tested the changes also in Firefox and Chrome). I can't assume every browser will be equipped with Intl, even though I've been told it's quite common these days

modules/sort/test/spec.coffee
jonthans: Apparently, that behaviour is consistent, after running the same test suite twice. It wasn't consistent before and probably isn't completely consistent now since James hinted that Intl.Collator tries to be quite clever. But I think for now it is good enough and at least now we can test new cases that come up

georgesimms: Is there a danger of api function proliferation here ? Would an options object help(eg `hx.sort.compare({ locale: true })(0, 1)` ? And then deprecate hx.sort.localeCompare

modules/sort/main/index.coffee
```diff
@@ -1,36 +1,60 @@
...
-  localeCollator = if hasCollator
-    new Intl.Collator(locale, options).compare
-  else
-    (a, b) ->
-      a.localeCompare(b, locale, options)
+  localeCollator = localeCollatorFn(locale, options)
 
   (a, b) ->
-    if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
+    if a? and b? and not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
```
c-frater: Could we pull this logic out into a function to share it? We've got it in every function so it seems sensible

modules/sort/test/spec.coffee
c-frater: That change does make sense but would mean making every comparison with `hx.sort.compare()(a,b)` which breaks the current api - I don't mind doing it but I'm not sure how we'd do it 'safely - unless we had `hx.sort.compare(a,b,options)`?

Also, we could add the `nullsLast` option to the `localeCompare` function as it already has an options object which would at least solve that issue.

modules/sort/main/index.coffee
```diff
@@ -41,20 +41,16 @@ localeCollatorFn = (locale, options) ->
 # slower than compare but enforces locale comparison for browsers that
 # dont support Intl.Collator.
 localeCompare = (locale, options) ->
```
c-frater: Would it be sensible to use `options = {}` as the argument? Probably sensible to always merge into an object (essentially equivalent to having `options ?= {}` as the first line of the function

modules/sort/main/index.coffee
```diff
@@ -1,36 +1,56 @@
...
+  options = hx.merge.defined(options, { numeric: true })
 
-  localeCollator = if hasCollator
-    new Intl.Collator(locale, options).compare
+  localeCollator = localeCollatorFn(locale, options)
+
+  if options.nullsLast
+    nullsLastCollator(localeCollator)
   else
     (a, b) ->
```
c-frater: could we move this to `defaultCollator(localeCollator)` and then use it in the `compare`, `compareNullsLast` and here as it's the same logic?

modules/sort/main/index.coffee
```diff
@@ -1,36 +1,56 @@
...
 
-compare = (a, b) ->
-  if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
+nullsLastCollator = (collator) -> (a, b) ->
+  if a is b then 0
+  else if a is undefined then 1
+  else if b is undefined then -1
+  else if a is null then 1
+  else if b is null then -1
+  else if not isNaN(Number(a)) and not isNaN(Number(b)) then a - b
```
c-frater: `else defaultCollator(collator)`? It probably doesn't matter that the default collator also checks `a? and b?`

PR #397, opened 2017-04-19
#396 Add full width to PivotTable

### LINKED ISSUE #396: Pivot Table: Add support for .hx-table-full and .hx-table-centered as a constructor option
## Context
This is not strictly necessary, but it would make the API more consistent and less confusing, especially when used in combination with sticky table headers. In particular, it would make a lot of sense to apply .hx-table-full by default unless the user explicitly disables it

c-frater:
## Description

Added the `fullWidth` option to pivot tables to pass through to StickyTableHeaders.

## Motivation and Context

Resolves #396 

## Checklist:

### DIFF
--- a/modules/pivot-table/main/index.coffee
+++ b/modules/pivot-table/main/index.coffee
@@ -56,7 +56,8 @@ class PivotTable extends hx.EventEmitter
       cellRender: (data, element, isHead, column) ->
         hx.select(element).text(data)
       useResponsive: true
-      data: undefined
+      data: undefined,
+      fullWidth: false
     }, options)
 
     @_ = {}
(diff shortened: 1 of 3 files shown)


modules/sticky-table-headers/main/index.coffee
```diff
@@ -192,7 +192,7 @@ class StickyTableHeaders
     offsetHeight = 0
     offsetWidth = 0
 
-    if options.fullWidth
+    if options.fullWidth or table.classed('hx-table-full')
```
georgesimms: this may be necessary but it is certainly not sufficient, it would make more sense to do this where the options get created.

The main issue is

```
if options.fullWidth
  table.style('min-width', wrapperBox.width + offsetWidth - widthScrollbarOffset - 1 + 'px')
```

Also, I would prefer it if the fullWidth option were completely removed - not only is it undocumented, but you should always be able to work out from context whether to use it or not so it's a bit redundant.

modules/pivot-table/main/index.coffee
```diff
@@ -56,7 +56,8 @@ class PivotTable extends hx.EventEmitter
       cellRender: (data, element, isHead, column) ->
         hx.select(element).text(data)
       useResponsive: true
-      data: undefined
+      data: undefined,
+      fullWidth: false
```
georgesimms: this is fine for our usecase, but maybe it should be set to "true" ? As you almost always want it

c-frater: Setting it to true would be a 'breaking' change - New features should be disabled by default so they don't break existing uses, something this is liable to do (i.e. causing layouts to break)

Issue #102, opened 2016-01-13
Table/Label: label font size is hard to read inside tables


c-frater:
`.hx-table` has a font size of 0.9em which is causing the issue
labels can be increased in size inside tables or tables can have normal size font, not sure which method would be better

PR #403, opened 2017-05-05
401 data table advanced filter criteria

### LINKED ISSUE #401: DataTable: filter with advanced criteria
## Context

Currently when using the advanced search it is only possible to do `equals/contains` filtering. It would be nice to add in greater than, less than, doesn't contain etc. filter types.

This should probably be under `advancedSearchCriteria` and should only be applied to the advanced search. 

Applying it to the 'regular' filter would probably require the api to be changed as the current behaviour is to convert all the cell values to a single string and use a 'contains' filter for the search term. The filter types other than 'contains' would not work in this case without changing the api for the `filter` function in the feeds. 

The UI should probably just be a `Picker` with the different  [Filter](https://www.hexagonjs.io/docs/filter/) types available

c-frater:
## Description

- Added `hx.filter.types`, `hx.filter.stringTypes` and `hx.filter.numberTypes` to expose the lists of filters for other modules to use
- Added the ability to use the `hx.filter` types when using the advancedSearch
- Changed the UI to make the advanced search flow better on smaller devices

#### API Changes
```js
hx.DataTable::advancedSearchCriteria([subset of hx.filter.types]) // Set global
hx.DataTable::advancedSearchCriteria(columnId, [subset of hx.filter.types]) // Set column
```
Added `criteria` argument to `termLookup` function (defaults to `contains` for the defaultTermLookup)

## Motivation and Context

Resolves #401 

## How Has This Been Tested?

TODO

#### Compact Flow Any Column
Compact Flow Any Column

#### Compact Flow Column
Compact Flow Column

#### Full Table Any Column
Full Table Any Column

#### Full Table Column
Full Table Column

#### Full Table Flow Column
Full Table Flow Column

## Checklist:

### DIFF
--- a/modules/data-table/main/index.coffee
+++ b/modules/data-table/main/index.coffee
@@ -289,7 +289,6 @@ class DataTable extends hx.EventEmitter
       showSearchAboveTable: false
       filter: undefined
       filterEnabled: true
-      filterCriteria: 'contains'
       showAdvancedSearch: false
       advancedSearchEnabled: false
       advancedSearchCriteria: undefined
@@ -1195,9 +1194,9 @@ objectFeed = (data, options) ->
         filterCacheTerm = range.advancedSearch
         sorted = undefined
       else
-        filterFn = (row) -> options.filter(range.filter, row, range.filterCriteria)
+        filterFn = (row) -> options.filter(range.filter, row)
         filterCache = getFiltered(data.rows, range.filter, filterCache, filterCacheTerm, filterFn)
-        filterCacheTerm = range.filter + (range.filterCriteria or '')
+        filterCacheTerm = range.filter
         sorted = undefined
 
       if sorted is undefined or sortCacheTerm.column isnt range.sort?.column or sortCacheTerm.direction isnt range.sort?.direction
--- a/modules/data-table/main/index.coffee
+++ b/modules/data-table/main/index.coffee
@@ -13,10 +13,17 @@ hx.userFacingText({
     rowsPerPage: 'Rows Per Page',
     search: 'Search',
     selectedRows: '$selected of $total selected.',
-    sortBy: 'Sort By'
+    sortBy: 'Sort By',
+    contains: 'contains',
+    excludes: 'does not contain',
+    startsWith: 'starts with',
+    fuzzy: 'fuzzy matches',
+    regex: 'matches regex',
+    exact: 'is exactly',
+    greater: 'is greater than',
+    less: 'is less than'
   }
 })
-
 
 fullWidthColSpan = 999 # the colspan used to make a cell display as an entire row
 collapseBreakPoint = 480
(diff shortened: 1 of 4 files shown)


modules/data-table/main/index.coffee
```diff
@@ -27,6 +34,13 @@ columnOptionLookup = (options, name, id) ->
   else
     options[name]
 
+toCriteriaItems = (list) ->
+  new hx.Set(list).values().map (item) ->
```
jmsmyth: Could also do `hx.unique(list).map (item) ->`

c-frater: Wasn't aware of `hx.unique` - Missed it because it's so close to the bottom of the [Util](https://www.hexagonjs.io/docs/util/) api section...

modules/data-table/main/index.coffee
```diff
@@ -168,42 +222,56 @@ createAdvancedSearchView = (selection, dataTable, options) ->
...
-      .component().value(if index is 0 then 'or' else 'and')
+    filterRowSel.select('.hx-data-table-advanced-search-type').component()
+      .value(if index is 0 then 'or' else 'and')
+
+    trueColumn = column or 'any'
+    filterRowSel.select('.hx-data-table-advanced-search-column').component()
+      .value(trueColumn)
+
+    columnCriteria = columnOptionLookup(options, 'advancedSearchCriteria', column) || []
+    criteriaItems = if trueColumn is 'any' then ['contains'] else ['contains', advancedSearchCriteriaValidate(columnCriteria)...]
```
c-frater: I couldn't manage to get the `set` validation to work as expected because it treats `advancedSearchCriteria('something')` as `get value for column 'something'` so instead I've applied the validation logic when rendering the table - The value can be set to an invalid value but won't be rendered and will show a console warning

c-frater:
Docs are in [Site #102](https://github.com/ocadotechnology/hexagonjs-site/pull/120)

modules/data-table/main/index.coffee
```diff
@@ -27,6 +34,25 @@ columnOptionLookup = (options, name, id) ->
...
+toCriteriaItems = (list) ->
+  hx.unique(list).map (item) ->
+    {
+      value: item,
+      text: hx.userFacingText('dataTable', item)
+    }
+
+advancedSearchCriteriaValidate = (value) ->
+  allowedTypes = hx.filter.types()
+  if (hx.isArray(value) and value.every((c) -> ~allowedTypes.indexOf(c))) or value is undefined
```
jmsmyth: Can this be changed to `>=0` or `!==-1` instead of tilde? More readable in my opinion

modules/data-table/main/index.coffee
```diff
@@ -27,6 +34,25 @@ columnOptionLookup = (options, name, id) ->
...
+    value or []
+  else if hx.isArray(value)
+    invalidTypes = value.filter((c) -> not ~allowedTypes.indexOf(c))
```
jmsmyth: Same thing with the `~` here

c-frater: Sure - Noticed your comments after I hit merge....
I'll fix it on master

PR #408, opened 2017-07-14
Add Selection::replace

jmsmyth:
## Description
Adds a new method that removes a selection from the DOM and replaces it with new content

## Motivation and Context
Makes it possible to do things like: 

    hx.select('body')
       .add(hx.spinner().replace(api.getDasboardData().then(dashboardComponent)))

So that a loading spinner is shown whilst the data is being fetched, and gets replaced with the `dashboardComponent` once the promise resolves.

`Selection::replace` also works with non-promise content:

    hx.select('.content').replace(hx.div('new-content'))

## How Has This Been Tested?
New unit tests have been added 

## Checklist:

### DIFF
--- a/modules/selection/main/index.coffee
+++ b/modules/selection/main/index.coffee
@@ -281,6 +281,20 @@ class Selection
       this.clear().add(children)
     return this
 
+  # replaces this selection with some other content
+  replace: (content) ->
+    # The use of Promise.resolve can delay when the replacement happens, so
+    # check to only do this when needed. This makes testing nicer when adding
+    # non-promise content
+    if content.then
+      Promise.resolve(content).then (sel) =>
+        this.insertAfter(sel)
+        this.remove()
+    else
+      this.insertAfter(content)
+      this.remove()
+    return this
+
   # gets the nth node in the selection, defaulting to the first
   node: (i=0) -> @nodes[i]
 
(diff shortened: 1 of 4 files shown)


modules/selection/main/index.coffee
```diff
@@ -271,7 +271,29 @@ class Selection
     this
 
   # clears the contents of a node and then adds the children passed in
-  set: (children) -> this.clear().add(children)
+  set: (children) ->
+    # The use of Promise.resolve can delay when the replacement happens, so
+    # check to only do this when needed. This makes testing nicer when adding
+    # non-promise content
+    if children.then
+      Promise.resolve(children).then((sel) => this.clear().add(sel))
```
c-frater: Should we use this promise logic for add/append chaining too? (i.e. the rest of the `Selection` module)

jmsmyth: I was wondering the same, but it's not so easy to do for the `add`/`append` methods since the order in which the element is added would depend on when the promises resolve. 

For `set` its easy to use promises since all the elements are being replaced - so there are no ordering issues. 

For `replace` it is also okay since you know the position of the currently selected element, so at the point when you want to perform the replace, the position to insert the new content can easily be found.

c-frater: Makes sense - Could we use `Promise.all` (to fix the order) and then `set` inside `add`/`append` in the case we're using promises?

Would be nice to be consistent is all 😄

PR #413, opened 2017-08-25
New demo page

c-frater:
Update to the demo page to have all the examples from the CSS / CSS + JS modules as well as some more UI examples to provide a single page with all of hexagon and allow for easier theming

The theory is that this would replace the current 'demo' index page - I've put it in the 'full' directory in case there's a reason this wouldn't be a great idea

*TODO*: 
- [x] Data tables
- [x] Plot
- [x] Titlebar

~- [ ] Drawing (not sure if we want / need this on here?)~ Not affected by theme and slows page down so not adding drawing example

*Screenshot*:
Screenshot


c-frater:
@jmsmyth Are you ok with me replacing the current 'demo' page with this one?

jmsmyth:
@c-frater Sounds fine to me

Issue #418, opened 2017-10-03
FastClick/DataTable: Can't click links in tables on iPad


c-frater:
## Context

When creating a link in a DataTable it isn't possible to click the link on an iPad. It looks like the data table row is consuming the event before it can reach the link - iPads don't seem to follow the standard event bubbling behaviour so it may be that the event is captured before the 'click' makes it to the hyperlink

## Expected Behavior

The link click would behave as normal

## Actual Behavior

The link click does not work - the user is not taken to another page

## Possible Fix

This may be fixed by adding `A` to the list of `DONT_PREVENT_DEFAULT_NODES` in the fast click module

## Steps to Reproduce

```js
const hx = window.hx

const exampleTable = hx.dataTable({
  feed: hx.dataTable.objectFeed({
    headers: [
      {name: 'Name', id: 'name'},
      {name: 'Age', id: 'age'},
      {name: 'Profession', id: 'profession'}
    ],
    rows: [
      {
        id: 0, // hidden details can go here (not in the cells object)
        cells: {'name': 'Bob', 'age': 25, 'profession': 'Developer'}
      },
      {
        id: 1,
        cells: {'name': 'Jan', 'age': 41, 'profession': 'Artist'}
      },
      {
        id: 2,
        cells: {'name': 'Dan', 'age': 41, 'profession': 'Builder'}
      }
    ]
  }),
  columns: {
    name: {
      cellRenderer: (e, d) => hx.select(e)
        .add(hx.detached('a').attr('href', 'http://www.google.co.uk').text(d))
    }
  }
})

hx.select('body')
  .clear()
  .add(exampleTable)
```
## Your Environment

* Browser Name and version:  Safari/Chrome
* Operating System and version (desktop or mobile): iPad (mobile)

PR #419, opened 2017-10-03
Potential Fix: Fast click handling for links

c-frater:
## Description

The current fast click handling causes links to be disabled in tables
as iPad event bubbling happens in a different order from desktops /
other devices.

Adding `A` to the list of elements seems to resolve the issue, however
it may be that it was intentionally not added to the list of elements.

## Motivation and Context

Related to: #418 

## How Has This Been Tested?

Tested in Chrome iPad emulator (will test on iPad when available)

## Checklist:

### DIFF
--- a/modules/fast-click/main/index.coffee
+++ b/modules/fast-click/main/index.coffee
@@ -4,7 +4,8 @@ if hx.supports('touch')
     'INPUT',
     'TEXTAREA',
     'SELECT',
-    'LABEL'
+    'LABEL',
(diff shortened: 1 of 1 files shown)


c-frater:
I think it's because the event bubbling means the row click triggers the fast click which then captures the link click. Testing with the change there was the 300ms delay before the link action happened so I'd expect it's just catching it before it finishes firing the lowest-level event on the link

PR #426, opened 2017-10-13
Titlebar: Add ability to set the active state using Node/Selection

### LINKED ISSUE #390: Titlebar: not possible to set active using an element
## Context

For a given link, it is not possible to set directly using the element, instead it expects a number or an ID selector

## Expected Behavior

`hx.titlebar.active(selection)` should set the active link to the selection (assuming it exists in the titlebar)

## Actual Behavior

It throws an error as `selection.node(id)` is called with the passed in selection

## Steps to Reproduce
On https://www.hexagonjs.io/docs/titlebar/
```js
const sel = hx.select('.example-heading')
sel.api().active(sel.selectAll('.hx-titlebar-link').node(2))
```

## Workaround
Add an ID to the node and use:
```js
hx.titlebar.active('#' + selection.attr('id'))
```

c-frater:
## Description

Resolves #390 

## Checklist:

### DIFF
--- a/modules/titlebar/main/index.coffee
+++ b/modules/titlebar/main/index.coffee
@@ -68,7 +68,6 @@ class TitleBar
     if arguments.length > 0
       selection = hx.selectAll('.hx-titlebar-link').classed('hx-selected', false)
       if id?
-        console.log('id', id, hx.select(id))
         @_.active = if hx.isString(id) or isElement(id) or isSelection(id)
           hx.select(id).classed('hx-selected', true)
         else
--- a/modules/titlebar/main/index.coffee
+++ b/modules/titlebar/main/index.coffee
@@ -23,6 +23,10 @@ setVisibility = (show, animate=true) ->
   animateTitlebar(hx.select(@selector).selectAll('.hx-titlebar-linkbar'))
 
 
+isElement = (obj) -> !!(obj and obj.nodeType is 1)
+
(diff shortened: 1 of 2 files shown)


modules/titlebar/main/index.coffee
```diff
@@ -23,6 +23,10 @@ setVisibility = (show, animate=true) ->
   animateTitlebar(hx.select(@selector).selectAll('.hx-titlebar-linkbar'))
 
 
+isElement = (obj) -> !!(obj and obj.nodeType is 1)
+
+isSelection = (obj) -> obj instanceof hx.Selection
```
c-frater: @jmsmyth Was going to add these to the util module but wasn't sure if it made sense?

Borrowed `isElement` from [Underscore.js](http://underscorejs.org/#isElement)

jmsmyth: `isSelection` has also been added in the webpack branch, but in the selection module, so would make sense to add it there if you want to add it to the public api now. Otherwise you could leave it here for now, and when this change gets merged with the webpack branch it can be deduplicated

For `isElement`, there is now a 'dom-utils' module in the webpack branch which would be a good place to put it, or maybe it could also go in the selection module. Probably okay to leave this one here for now

c-frater: Ok, I'll leave them both here and raise a ticket to move them once webpack is merged

PR #436, opened 2018-04-16
Fix the 'edit page' links

### LINKED ISSUE #437: Docs: The edit page links are broken
Go to http://hexagonjs.io/docs/base/ and click on the 'edit page' link in the top right corner. The link directs to the wrong place because it still references 'hexagonjs-site' and the `{{filename}}` variable hasn't been set up in the quantum config

jmsmyth:
This PR fixes the docs so that the edit page links work

Fixes #437


docs/quantum.config.js
```diff
@@ -84,7 +84,8 @@ function customizedTemplate (file) {
     variables: {
       baseurl: baseUrl,
       version: file.meta.version,
-      latestVersion: latestVersion
+      latestVersion: latestVersion,
+      editPageUrl: `https://github.com/ocadotechnology/hexagonjs/tree/master/docs/${file.info.src}`
```
c-frater: +1 for moving the whole URL into a variable

PR #443, opened 2018-05-31
fix(builder): Resolve issue with `moduleDirectory` option

c-frater:
## Description

The option allows an additional directory to be passed in to extend
hexagon however part of the compiler config only allowed for modules
to be added from the root directory of core hexagon, causing the builder
to break when passing in `moduleDirectory`

## Motivation and Context

Trying to add an internal-only module to our Hexagon build

## How Has This Been Tested?

Manual tests to ensure build output is unchanged when not using the `moduleDirectory` option.

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "hexagon-js",
-  "version": "1.14.0",
+  "version": "1.15.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1301,7 +1301,7 @@
     },
     "event-stream": {
       "version": "3.3.4",
-      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
+      "resolved": "http://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
```
jmsmyth: How come this changed?

c-frater: I've got no idea - I'll push an updated version

PR #499, opened 2018-07-11
fix(plot): Add defaults for `x` and `y` property for series label formatters

### LINKED ISSUE #348: Plot: labelFormatter issues with graphs
## Context

When creating a Bar series with `y` values over 100, the values are formatted by `hx.format.si(2)` (which is the default `axis` formatter) as defined by the default `labelValuesExtractor`

It was expected that setting `labelFormatters.y` when adding a series would override that formatter but it is not applied, making it impossible to set the label formatter correctly for bar charts without providing a new `labelValuesExtractor`.

The default label formatter should also probably be `hx.identity` rather than the axis formatter as it should normally show the actual value. 

## Expected Behavior

```js
axis.addSeries('bar', {
 ...
 labelFormatters: {
   y: function...
   x: function...
 }
})
```
Should override the default label formatter

## Actual Behavior

The passed in label formatter is ignored and the value is formatted with `hx.format.si(2)` (the axis formatter)

## Possible Fix

The following would probably resolve all the issues in this area:
- Update `Series` to have defaults for `labelFormatters.x/y` 
- Update the default `labelValuesExtractor` to fall back to `hx.identity` instead of `series.axis.y.formatter()`
- Update the`xProperty` and `yProperty` in the default `labelValuesExtractor` to have defaults of `'x'` and `'y'`

## Steps to Reproduce

```js
const graphSel = hx.select('body').append('div')
const graph = new hx.Graph(graphSel.node())
const axis = graph.addAxis({
  x: {
    scaleType: 'discrete'
  },
  y: {
    min: 0
  }
})


axis.addSeries('bar', {
  title: 'Bar Series 2',
  fillColor: hx.theme.plot.colors[3],
  data: [{ x: 0, y: 111 }],
  labelFormatters: {
    y: hx.identity
  }
})

graph.render()
```
Value in label is rendered as `110` instead of `111`

## Workaround
For graphs with sensible values (that don't need `si` formatting), the `axis.x/y.formatter` can be set to `hx.identity`, e.g.:
```js
axis.x.formatter(hx.identity)

graph.addAxis({
  x: {
    formatter: hx.identity
  }
})
```

## Your Environment

* Hexagon Version: 1.8.2

c-frater:
## Description

Added default `xProperty` and `yProperty` to the `defaultLabelValuesExtractor` 

## Motivation and Context

Resolves #348 

## How Has This Been Tested?

Tested in browser

screen shot 2018-07-11 at 12 34 37

## Checklist:

### DIFF
--- a/modules/plot/main/coffee/Series.coffee
+++ b/modules/plot/main/coffee/Series.coffee
@@ -1,7 +1,7 @@
 
 class Series extends hx.EventEmitter
 
-  defaultLabelValuesExtractor = (series, dataPoint, xAccessor, yAccessor, xProperty, yProperty) ->
+  defaultLabelValuesExtractor = (series, dataPoint, xAccessor, yAccessor, xProperty='x', yProperty='y') ->
     [
       {
         name: series.axis.x.title(),
--- a/docs/content/docs/plot/api/auto.um
+++ b/docs/content/docs/plot/api/auto.um
@@ -36,6 +36,12 @@
   @description
     Prevent text wrapping in basic labels (e.g. sparklines)
 
+@bugfix nextReleaseVersion
+  @issue 348
+  @description
+    Resolves an issue where the label formatters were not correctly being read
+    for series.
+
 @object hx.theme.plot
   @property colors [Array[String]]
     @description


modules/plot/main/coffee/Series.coffee
```diff
@@ -1,7 +1,7 @@
 
 class Series extends hx.EventEmitter
 
-  defaultLabelValuesExtractor = (series, dataPoint, xAccessor, yAccessor, xProperty, yProperty) ->
+  defaultLabelValuesExtractor = (series, dataPoint, xAccessor, yAccessor, xProperty='x', yProperty='y') ->
```
c-frater: After chasing this down, I could only see [one place](https://github.com/ocadotechnology/hexagonjs/blob/1-17-0/modules/plot/main/coffee/Utils.coffee#L272) where these values are actually passed through by the code. This meant that all series using the default were doing `series.labelFormatter(undefined)` instead of `series.labelFormatter('x')` and essentially ignoring the `labelFormatters` option altogether

PR #501, opened 2018-07-11
fix(autocomplete-picker): Update change event to emit on undefined value

### LINKED ISSUE #438: AutocompletePicker: No events fired when clearing the value
## Context

When setting the AutocompletePicker value to `undefined` the `change` event is not fired

## Possible Fix

Updating the [`setPickerValue`](https://github.com/ocadotechnology/hexagonjs/blob/master/modules/autocomplete-picker/main/index.coffee#L22-L31) function to emit when there both is and is not a value

c-frater:
## Description

Changed the event to be emit when setting the value to undefined

## Motivation and Context

Resolves #438 

## How Has This Been Tested?

Added regression test

## Checklist:

### DIFF
--- a/docs/content/docs/autocomplete-picker/api/AutocompletePicker/events/change.um
+++ b/docs/content/docs/autocomplete-picker/api/AutocompletePicker/events/change.um
@@ -1,4 +1,10 @@
 @event change [String]
+  @bugfix nextReleaseVersion
+    @issue 438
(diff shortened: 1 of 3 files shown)


c-frater:
I didn't realise until after raising this but this PR depends on #502 as without it you cannot set the value to undefined (and the test fails)

PR #519, opened 2018-12-11
Paginator V2 Update

c-frater:
## Description

Update the paginator component to add a more accessible rendering style that uses `aria-{xxx}` and provides support for navigation using the keyboard as documented [here](http://www.a11ymatters.com/pattern/pagination/) with the addition of a prev/next button

## Motivation and Context

Resolves some concerns that were raised with the existing paginator component and aligns more with the standard for pagination.

## How Has This Been Tested?

Added tests for the logic for rendering the buttons and tested in the browser. 

Screenshot generated based on specs passed to tests
Screenshot

## TODO
- [x] Tests for the rendered component
- [x] Validate that the existing component has not changed behaviour (as it had no tests)
- [x] Docs

## Checklist:


modules/paginator/main/index.coffee
```diff
@@ -1,133 +1,320 @@
+hx.userFacingText({
+  paginator: {
+    paginatorAria: 'Pagination navigation',
+    currentPageAria: 'Current page, page $page',
+    gotoPageAria: 'Goto page $page',
+    prevPageAria: 'Goto previous page, page $page',
+    nextPageAria: 'Goto next page, page $page',
+    prev: 'Prev',
+    next: 'Next',
```
lordmat0: How would we pass in different languages?

c-frater: Two methods, one is to pass to the constructor (which is where these values are consumed) on a per-component basis or you can override `hx.userFacingText`:
```
new Paginator('#selector', {
  prevText: 'Bob!'
})
```

```
hx.userFacingText({
  paginator: { prev: 'Bob!' }
})

new Paginator('#selector') // Uses `Bob!`
```

PR #529, opened 2019-05-24
chore(package): Remove package.main as hexagon is not for `node.js`

c-frater:
## Description

- Removed `package.main`
- Require package.json to get `package.module` for the output for the `es.js` bundle
- Add module bundle to dev/watch script

## Motivation and Context

The package.main was incorrectly using `hexagon.core.js` (an esm output)

As this package is not intended for use within `node.js` we don't need
to have a `package.main` or a `commonjs` output

## Checklist:

- [x] I have read the **CONTRIBUTING** document. (Last updated 18th May 2019)


c-frater:
After some discussion, we realised that some test frameworks (like `jest`) need `package.main` to be able to include components as they are running in `node.js` as opposed to a real browser.

Without this, you have to hack around with config or pre-compile hexagon using something like `babel` explicitly (as `node_modules` are often excluded by default)

Adding a `commonjs` package.main that is equivalent to the `package.module` (esm) is the required solution

PR #538, opened 2019-07-18
fix: update dropdown position calculation

c-frater:
## Description

Applies a directional class to the dropdown to apply the margin spacing on the correct side of the dropdown and resolves an issue where the dropdown could grow and cover the selection by switching to use `bottom` position instead of `top` when shown above the selection.

## Motivation and Context

Currently, the dropdown applies a top margin in the styles to add spacing between the dropdown and the selection. 

When the dropdown is in any position other than 'down' (i.e. below the selection) this top margin either causes the dropdown to be slightly misaligned or to overlap with the content in the dropdown.

This change prevents the dropdown overlapping when the spacing is greater than zero.

## How Has This Been Tested?

- Added tests to the check the alignment works correctly.
- All existing tests pass.

Before(5px dropdown spacing)         After
Screenshot 2019-07-18 at 15 14 20 Screenshot 2019-07-18 at 15 13 11

## Checklist:

- [x] I have read the **CONTRIBUTING** document. (Last updated 18th May 2019)

### DIFF
--- a/src/components/dropdown/index.coffee
+++ b/src/components/dropdown/index.coffee
@@ -15,7 +15,7 @@ export config = {
 checkFixedPos = (node) ->
   if select(node).style('position') is 'fixed' then true
 
-positionDropdown = ({alignments, selection, dropdown, useScroll}, {matchWidth}) ->
+positionDropdown = ({selection, dropdown, useScroll}, {align, matchWidth}) ->
   dropdown.style('display', 'block')
 
   # extract measurements from the dom
@@ -26,18 +26,23 @@ positionDropdown = ({alignments, selection, dropdown, useScroll}, {matchWidth}) 
   zIndex = parentZIndex(selection.node(), true)
 
   # calculate the position of the dropdown
-  {x, y} = calculateDropdownPosition(
-    alignments,
+  {x, y, direction} = calculateDropdownPosition(
+    dropdown.attr('data-direction') or align,
     { x: rect.left, y: rect.top, width: rect.width, height: rect.height },
     { width: dropdownRect.width, height: dropdownRect.height },
     { width: window.innerWidth, height: window.innerHeight },
     ddMaxHeight,
-    scrollbarSize()
+    scrollbarSize(),
   )
 
+  yPos = 'top'
   if not parentFixed
     x += window.scrollX || window.pageXOffset
     y += window.scrollY || window.pageYOffset
+
+    if direction is 'up'
+      yPos = 'bottom'
+      y = document.body.clientHeight - y - dropdownRect.height
 
   # update the styles for the dropdown
   if zIndex > 0
@@ -53,8 +58,15 @@ positionDropdown = ({alignments, selection, dropdown, useScroll}, {matchWidth}) 
     dropdown.style('overflow-y','auto')
 
   dropdown
+    .classed('hx-dropdown-up', direction is 'up')
+    .classed('hx-dropdown-down', direction is 'down')
+    .classed('hx-dropdown-left', direction is 'left')
+    .classed('hx-dropdown-right', direction is 'right')
+    .attr('data-direction', direction)
+    .style('top', 'auto')
+    .style('bottom', undefined)
+    .style(yPos, y + 'px')
     .style('left', x + 'px')
-    .style('top', y + 'px')
 
 dropdownContentToSetupDropdown = (dropdownContent) ->
   # XXX Breaking: Renderer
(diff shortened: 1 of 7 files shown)


src/components/dropdown/index.coffee
```diff
@@ -26,19 +26,24 @@ positionDropdown = ({alignments, selection, dropdown, useScroll}, {matchWidth})
   zIndex = parentZIndex(selection.node(), true)
 
   # calculate the position of the dropdown
-  {x, y} = calculateDropdownPosition(
-    alignments,
+  {x, y, direction} = calculateDropdownPosition(
+    dropdown.attr('data-direction') or align,
```
c-frater: Storing the `data-direction` means the dropdown doesn't 'jump around'.

This solves an issue specific to things like autocomplete pickers.

- Open autocomplete picker with 10 items near the bottom of the page (i.e with space for only 5 items to show in the dropdown)
    - Dropdown shown above button
- Enter value in the filter to reduce size of dropdown (so it would fit below)
    - Dropdown jumps to below selection

After the change, the dropdown will continue to display above the selection (unless there isn't room, in which case it would jump below the selection again)

src/components/dropdown/index.coffee
```diff
@@ -26,19 +26,24 @@ positionDropdown = ({alignments, selection, dropdown, useScroll}, {matchWidth})
...
   )
 
+  yPos = 'top'
   if not parentFixed
     x += window.scrollX || window.pageXOffset
     y += window.scrollY || window.pageYOffset
 
+    if direction is 'up'
+      yPos = 'bottom'
+      y = document.body.clientHeight - y - dropdownRect.height
```
c-frater: Taken from the old hexagon code. I'm not sure when this logic was removed (or why) but it's mostly required for browsers that don't support the `MutationObserver` (as otherwise the increase/decrease in size is applied from the top of the dropdown, causing an overlap with the selection)

src/components/dropdown/index.coffee
```diff
@@ -53,8 +58,15 @@ positionDropdown = ({alignments, selection, dropdown, useScroll}, {matchWidth})
     dropdown.style('overflow-y','auto')
 
   dropdown
+    .classed('hx-dropdown-up', direction is 'up')
+    .classed('hx-dropdown-down', direction is 'down')
+    .classed('hx-dropdown-left', direction is 'left')
+    .classed('hx-dropdown-right', direction is 'right')
+    .attr('data-direction', direction)
+    .style('top', 'auto')
+    .style('bottom', undefined)
```
c-frater: We need `top` of `auto` to override the CSS style of `-10000px` and `bottom` of `undefined` to clear the bottom value when switching between `up` and `down`

src/components/dropdown/positioning.coffee
```diff
@@ -1,6 +1,14 @@
...
-export calculateDropdownPosition = (alignments, selectionRect, dropdownRect, windowRect, ddMaxHeight, scrollbarWidth) ->
+export calculateDropdownPosition = (align, selectionRect, dropdownRect, windowRect, ddMaxHeight, scrollbarWidth) ->
+  alignQuad = switch align
+    when 'up' then 'ltlb'
+    when 'down' then 'lblt'
+    when 'left' then 'ltrt'
+    when 'right' then 'rtlt'
+    else align
+
+  alignments = alignQuad.split('')
```
c-frater: Moved this inside the `position` calculation so this function now takes the `align` string as passed into the dropdown. This also means we can use the `direction` to save the position

src/components/dropdown/spec.coffee
```diff
@@ -106,24 +105,6 @@ export default () ->
...
-      dd._.alignments.should.eql('ltlb'.split(''))
-
-      dd = new Dropdown(button, content, {align: 'down'})
-      dd._.alignments.should.eql('lblt'.split(''))
-
-      dd = new Dropdown(button, content, {align: 'left'})
-      dd._.alignments.should.eql('ltrt'.split(''))
-
-      dd = new Dropdown(button, content, {align: 'right'})
-      dd._.alignments.should.eql('rtlt'.split(''))
```
c-frater: `dd._.alignments` is no longer stored as we work it out using the option directly each time the dropdown is rendered

src/components/form/theme.css
```diff
@@ -152,6 +152,10 @@
   height: var(--input-height);
   line-height: var(--input-line-height);
   font-size: 1em;
+}
+
+.hx-flag-form .hx-date-time-picker .hx-date-picker,
+.hx-flag-form .hx-date-time-picker .hx-time-picker {
   margin: calc(-1 * var(--input-border-width));
```
c-frater: Fixed a small issue with the date time picker spacing being different from the other dropdown spacing

src/hexagon.variables.css
```diff
@@ -389,7 +389,7 @@
   --dropdown-background-color: var(--theme-container-background-color);
   --dropdown-border-color: var(--theme-container-border-color);
   --dropdown-shadow-color: var(--theme-shadow-color);
-  --dropdown-spacing: 0;
+  --dropdown-spacing: 0.2em;
```
c-frater: The `0` spacing looked odd, I believe it was added because non-zero values didn't work (i.e. because of the issue this PR solves)

PR #544, opened 2019-09-02
Feature/visualisation bar

fiWhy:
#Visualisation bar: implemented visualisation bar

## Description
Created visualisation bar component with fake animation

## Motivation and Context
Task

## How Has This Been Tested?
No tests provided yet

## Checklist:
- [x] I have read the **CONTRIBUTING** document. (Last updated 16 Aug 2019)
- [ ] This request is ready to review and merge


src/components/visualisation-bar/index.js
```diff
@@ -0,0 +1,246 @@
...
+        this.buffer = this.options.buffer;
+        this.content = this.options.content;
+        this.disabled = this.options.disabled;
+        this.valid = this.options.valid;
+
+        if (this.options.mock && this.options.valid && !this.options.disabled) {
+            setInterval(() => {
+                this.progress = Math.round(Math.random() * 20);
+            }, 1000);
+        }
```
c-frater: This option and code should not be in the component. Updating based on an interval should live in the Demo or docs instead

src/components/visualisation-bar/index.js
```diff
@@ -0,0 +1,246 @@
...
+        this.countElement.text(`${this.options.value}/${this.options.max || '-'}`);
+    }
+}
+
+
+export const visualisationBar = (options) => {
+    const selection = div();
+    new VisualisationBar(selection.node(), options);
+    return selection;
+}
```
c-frater: I can see quite a few linting issues here

You should be able to run `npm run lint -- --fix` to solve quite a few of them, it will list the remaining ones to fix for you

src/demo/examples/visualisation-bar/index.js
```diff
@@ -0,0 +1,10 @@
+
+import { visualisationBar } from 'hexagon-js';
+
+export default () => [
+  visualisationBar({ title: 'With maximum value static', buffer: 2, value: 1, max: 20, mock: false }),
+  visualisationBar({ title: 'With disabled state', buffer: 2, value: 1, max: 20, disabled: true, content: 'Hello world!' }),
+  visualisationBar({ title: 'With maximum value', buffer: 2, value: 5, max: 20, mock: true }),
+  visualisationBar({ title: 'Without maximum value', buffer: 2, value: 5, mock: true, withPercent: true }),
+  visualisationBar({ title: 'With content', buffer: 2, value: 5, mock: true, withPercent: true, content: 'Hello world!' })
```
c-frater: Could you add a screenshot of these demos to the PR description?

src/hexagon.core.js
```diff
@@ -149,6 +149,7 @@ export { inputGroup } from 'components/input-group';
 export { group, section } from 'components/layout';
 export { dragContainer, DragContainer } from 'components/drag-container';
 export { progressBar, ProgressBar } from 'components/progress-bar';
+export { visualisationBar, VisualisationBar } from 'components/visualisation-bar';
```
c-frater: The tests will fail due to these exports. Need to add them to `hexagon.spec.js` (there's a big list of exported members in the `hx export the right things` test)

src/components/visualisation-bar/index.css
```diff
@@ -0,0 +1,150 @@
...
+  border-top-right-radius: var(--visual-border-radius);
+  border-bottom-right-radius: var(--visual-border-radius);
+}
+
+.without-progress .hx-visualisation-bar-buffer .hx-visualisation-bar-fill {
+  border-top-left-radius: var(--visual-border-radius);
+  border-bottom-left-radius: var(--visual-border-radius);
+}
+
+.without-balance .hx-visualisation-bar-buffer .hx-visualisation-bar-fill {
```
c-frater: All classes should be prefixed with `hx-` and in the case of component classes with `hx-<component name>` (i.e. `hx-visualization-bar-...`)

src/components/visualisation-bar/index.js
```diff
@@ -0,0 +1,246 @@
...
+    [label, div(`hx-visualisation-bar-label hx-visualisation-bar-${title}-label`)],
+    [labelPc, span('hx-visualisation-bar-label-percent')],
+    [labelText, span('hx-visualisation-bar-label-text')],
+
+])
+
+/**
+ * 
+ */
+export class VisualisationBar {
```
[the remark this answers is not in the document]
fiWhy: Sorry, was copied from doc...

src/components/visualization-bar/index.js
```diff
@@ -0,0 +1,288 @@
...
+
+/**
+ * 
+ */
+export class VisualizationBar {
+    constructor(selector, options) {
+        this.percents = {
+            [BarTypes.PROGRESS]: 0,
+            [BarTypes.BUFFER]: 0,
+            [BarTypes.BALANCE]: 100
```
c-frater: From my understanding, this component only supports 3 segments. After speaking to UX, it seems the component should support `n` segments, meaning a lot of the code associated to these `BarTypes` is somewhat redundant

src/hexagon.core.js
```diff
@@ -63,7 +63,7 @@ export {
   shallowClone,
   vendor,
   identity,
-} from 'utils/utils';
+} from './utils/utils';
```
c-frater: Not sure why these paths were made into relative imports?

src/components/visualization-bar/index.js
```diff
@@ -0,0 +1,387 @@
...
+import { mergeDefined, defined } from 'utils/utils';
+import { EventEmitter } from 'utils/event-emitter';
+import {
+  VisualizationBarTypes, VisualizationBarParts, VisualizationBarSizes,
+  generateGroup, VisualizationBarSizesList, VisualizationBarTypesList,
+} from './lib';
+
+const moreThan = target => value => (value < target ? target : value);
+const lessThan = target => value => (value > target ? target : value);
+const adaptNumber = target => value => moreThan(target)(lessThan(0)(value));
```
c-frater: `hx.clamp(0, target, value)` would achieve the same result here

src/components/visualization-bar/index.css
```diff
@@ -0,0 +1,291 @@
...
+.hx-visualization-bar-count {
+  width: 30%;
+  flex-direction: row-reverse;
+}
+
+.hx-visualization-bar-fill-wrapper {
+}
+
+.hx-visualization-bar-fill {
+  background-color: var(--visual-healthy-na-color);
```
c-frater: I can still see quite a few variables used in `index.css` here, these styles should be declared inside `theme.css` (i.e. don't use variables inside `index.css`)

src/hexagon.variables.css
```diff
@@ -1085,14 +1043,12 @@
...
   --titlebar-title-text-color: var(--theme-accent-color);
 
-
   /* Toggle */
-  --toggle-background-off-color: #D5D5D5;
-  --toggle-background-on-color: #D5D5D5;
-  --toggle-toggle-off-color: #ABABAB;
+  --toggle-background-off-color: #d5d5d5;
+  --toggle-background-on-color: #d5d5d5;
+  --toggle-toggle-off-color: #ababab;
```
c-frater: Not sure why most of this file has been changed in this PR, most of the changes don't affect the output and removing the additional newlines between sections makes it more difficult to read

It's fine to make these kind of changes, just generally they are done in their own PR

src/hexagon.variables.css
```diff
@@ -1103,4 +1059,24 @@
...
+  --visual-content-margin: 15px;
+  --visual-font-size: var(--theme-font-size);
+  --visual-base-inner-padding: 5px;
+  --visual-small-font-size: var(--base-small-font-size);
+  --visual-small-content-margin: 10px;
+  --visual-border-radius: 3px;
+  --visual-text-color: #fff;
+  --visual-inversed-text-color: #424242;
+  --visual-disabled-text-color: #c9c9c9;
+  --visual-min-screen-width: 50px;
```
c-frater: These variables should be `--visualization-bar-xxx` for consistency with other modules

src/demo/examples/visualization-bar/index.js
```diff
@@ -0,0 +1,90 @@
...
+      max: 100,
+      withPercent: true,
+      content: 'Hello world!',
+      progressLabel: 'Progress',
+      bufferLabel: 'Buffer Long Text!',
+      balanceLabel: 'Balance',
+    }), rand(1000)],
+    [new VisualizationBar(selectionSmall.node(), {
+      size: VisualizationBarSizes.SMALL,
+      title: 'Small progress bar',
```
c-frater: The `ProgressBar` and `VisualizationBar` should probably be separate components to reduce the API complexity

They do share some common visuals so could both initialise a shared class that has all the functionality of both components

src/components/visualization-bar/spec.js
```diff
@@ -0,0 +1,19 @@
...
+    it('should create and class the progress bar correctly', () => {
+      const pb = new VisualizationBar(div());
+      pb.selection.classed('hx-visualization-bar').should.equal(true);
+      pb.selection.selectAll('.hx-progress-bar-inner').size().should.equal(1);
+      pb.progress.should.equal(0);
+      /*
+        This method is used rather than calculate the pixel width as .style('width')
+        returns Npx not N%
+      */
+      pb.innerBars.attr('style').split(' ').join('').should.equal('width:0%;');
```
c-frater: I'm moving gradually towards using [this brilliant JS testing guide](https://github.com/mawrkus/js-unit-testing-guide) so these tests will need updating

For instance, each expectation should be moved into its own test case:
```js
describe('when creating a bar with no options', () => {
  let thisSel;
  let thisVisualizationBar;

  beforeEach(() => {
    thisSel = div();
    thisVisualizationBar = new VisualizationBar(thisSel);
  });

  it('sets the class correctly', () => {
    thisSel.classed('hx-visualization-bar').should.equal(true);
  });

  it('creates a single bar', () => {
    thisSel.selectAll('.hx-progress-bar-inner').size().should.equal(1);
  });

  it('sets the progress to zero', () => {
    thisVisualizationBar.progress().should.equal(0);
  });

  it('sets the width of all bars to zero percent', () => {
    /*
    This method is used rather than calculate the pixel width as .style('width')
    returns Npx not N%
  */
  thisVisualizationBar._.innerBars.attr('style').split(' ').join('').should.equal('width:0%;');
  });
});
```

As a personal preference, I omit `should` from the start of each test case. There's no point in adding it everywhere as it just pollutes the test output. (i.e. `when creating a bar with no options > sets the class correctly` reads just as well as `when creating a bar with no options > should set the class correctly`)

src/demo/examples/visualization-bar/index.js
```diff
@@ -0,0 +1,90 @@
...
+  e.progress(randNum(max));
+};
+
+export default () => {
+  const selectionWithMax = div();
+  const selectionErrored = div();
+  const selectionWithoutMax = div();
+  const selectionWithContent = div();
+  const selectionWithLabels = div();
+  const selectionSmall = div();
```
c-frater: I'd just use the `div()` inline instead of creating these `const` values to keep the script tidy

src/demo/examples/visualization-bar/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+  const mocks = [
+    [new VisualizationBar(selectionWithMax.node(), {
+      title: 'With maximum value', buffer: 2, value: 5, max: 20,
+    }), rand(20)],
```
c-frater: If you're wrapping these in arrays, could you format it like:

```js
[
  new VisualizationBar(selectionWithMax.node(), {
    title: 'With maximum value', buffer: 2, value: 5, max: 20,
  }), 
  rand(20),
]
```

Makes it much easier to read. Without this, it's difficult to notice that `rand(20)` isn't part of the `VisualizationBar` construction)

c-frater:
Closing - Resolved in #547

PR #551, opened 2019-10-23
Feature/tooltip: Add tooltip function

kbberker:
## Description

Add a tooltip with either a label or icon. Alignment is currently MVP. 

## Motivation and Context

Added after UX requirements.

## How Has This Been Tested?

Unit tests have been written.
A demo has been added and tested in the browser.

Screenshot 2019-11-01 at 09 49 38

## Checklist:

- [x] I have read the **CONTRIBUTING** document. (Last updated 16 Aug 2019)


src/components/tooltip/index.css
```diff
@@ -0,0 +1,9 @@
+.hx-tooltip-label  {
+    border-bottom: 1px dotted red;
```
c-frater: I forgot to go over theming with you. We need to move everything that isn't part of the 'theme' (e.g. `display`, `overflow` stuff like that) into this file and add variables for everything else.

In this case in `index.css`
```css
.hx-tooltip-label {
  border-bottom: 1px dotted transparent;
}
```

Then inside `theme.css`
```css
.hx-tooltip-label {
  border-bottom-color: var(--tooltip-border-color);
}
```

Variable definitions then go into `hexagon.variables.css`. The file is in alphabetical order and there is a heading for each component/module

Something like this might suit:
```css
/* Tooltip */
--tooltip-border-color: var(--theme-accent-color);
```

src/components/tooltip/spec.js
```diff
@@ -0,0 +1,150 @@
...
+
+import installFakeTimers from 'test/utils/fake-time';
+import emit from 'test/utils/fake-event';
+
+import { tooltip } from 'components/tooltip';
+import { config as dropdownConfig } from 'components/dropdown';
+
+export default () => {
+  describe('tooltip', () => {
+    let clock;
```
c-frater: The dropdown no longer animates so we can remove all code associated with the `clock` here

src/components/tooltip/index.css
```diff
@@ -0,0 +1,21 @@
...
+.hx-tooltip-label  {
+    border-bottom: 1px dotted transparent;
+}
+
+.hx-tooltip-icon,
+.hx-tooltip-label {
+    cursor: pointer;
+}
+
+.hx-tooltip:before {
```
c-frater: Need to add support for when the tooltip is close to the top of the page (and the dropdown is shown below instead of above)

Issue #550, opened 2019-10-23
Dropdown: initial positioning sometimes displays incorrectly


c-frater:
## Context

When opening a dropdown for the first time it sometimes displays in an incorrect position.

## Steps to Reproduce

Unable to reliably reproduce, except with some screenshot testing (which is difficult to debug)

## Your Environment

* Hexagon Version: 2.4.2

p-koscielniak:
Hi
This issue is caused by using ```position: absolute; bottom: **px``` instead of ```position: absolute; top: **px``` for dropdown positioning.

`bottom`  positioning shouldn't be used because resizing window or changing viewport orientation changes dropdown position. 

Removing:
```
    if direction is 'up'
      yPos = 'bottom'
      y = document.body.clientHeight - y - dropdownRect.height
```
fixes flaky BackstopJS tests.

PR #556, opened 2019-12-03
Remove dropdown bottom positioning.

p-koscielniak:
Fix for https://github.com/ocadotechnology/hexagonjs/issues/550


src/components/dropdown/index.coffee
```diff
@@ -35,15 +35,10 @@ positionDropdown = ({selection, dropdown, useScroll}, {align, matchWidth}) ->
...
   )
 
-  yPos = 'top'
   if not parentFixed
     x += window.scrollX || window.pageXOffset
     y += window.scrollY || window.pageYOffset
 
-    if direction is 'up'
-      yPos = 'bottom'
-      y = document.body.clientHeight - y - dropdownRect.height
```
c-frater: This code is designed to allow the margins to be applied to space the dropdown correctly in relation to the dropdown. 

Before this change (and changing the `--dropdown-spacing` to 15px):

*Dropdown down*: - Spacing shown between button and dropdown
Screenshot 2019-12-04 at 15 02 51

*Dropdown up*: - Same dropdown, not enough space to show downwards so flip it upwards
Screenshot 2019-12-04 at 15 02 58


After the change:
Screenshot 2019-12-04 at 15 03 32

Essentially, as we apply `margin-bottom` it means the dropdown is now shown without spacing.

I'm sure there's a way to resolve this with changing the CSS, worth fiddling with it 😉

src/components/dropdown/theme.css
```diff
@@ -6,7 +6,7 @@
 }
 
 .hx-dropdown-up {
-  margin-bottom: var(--dropdown-spacing);
+  margin-bottom: calc(-1 * var(--dropdown-spacing));
```
c-frater: I believe this needs `margin-top` instead as `margin-bottom` has no effect when using `top: XXXpx`

p-koscielniak: Yes, margin-top is what I've tested locally. Creating commit via GitHub web GUI wasn't the best idea :)

